Issue #142, opened 2015-11-03
Incorrect results of matrix multiplication and matrix concatenation of Nd4j library


beomyeol:
We are using `0.0.3.5.5.5` version of [ND4J](http://nd4j.org/) library for matrix operations. However, I found that matrix multiplication results are incorrect when getting one of operands via `getRow()` method of `INDArray`. `getRow()` method is essential for batch processing.
In addition, the concatenation operation of matrices produces incorrect results of which shapes are wrong. In model partitioning, the concatenation operation is necessary.

With the latest version `0.4-rc3.5` of ND4J, the above issues seem to be fixed. But, multiplication results are still incorrect, which makes `NerualNetworkTest` fail. Although we have watched the updates of ND4J library for more than 2 months, the accuracy problem of matrix multiplication has not been fixed yet.

We may use workarounds for these. But it will make our codes complicated. Thus, we should consider the replacement of ND4J library with other stable, reliable, and faster BLAS libraries such as [JBLAS](http://http://jblas.org/) and [netlib-java](https://github.com/fommil/netlib-java).

bgchun:
+1 on using JBLAS and netlib-java.


On Tue, Nov 3, 2015 at 4:59 PM, Beomyeol Jeon <notifications@github.com>
wrote:

> We are using 0.0.3.5.5.5 version of ND4J <http://nd4j.org/> library for
> matrix operations. However, I found that matrix multiplication results are
> incorrect when getting one of operands via getRow() method of INDArray.
> getRow() method is essential for batch processing.
> In addition, the concatenation operation of matrices produces incorrect
> results of which shapes are wrong. In model partitioning, the concatenation
> operation is necessary.
>
> With the latest version 0.4-rc3.5 of ND4J, the above issues seem to be
> fixed. But, multiplication results are still incorrect, which makes
> NerualNetworkTest fail. Although we have watched the updates of ND4J
> library for more than 2 months, the accuracy problem of matrix
> multiplication has not been fixed yet.
>
> We may use workarounds for these. But it will make our codes complicated.
> Thus, we should consider the replacement of ND4J library with other stable,
> reliable, and faster BLAS libraries such as JBLAS
> <http://http://jblas.org/> and netlib-java
> <https://github.com/fommil/netlib-java>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/cmssnu/dolphin/issues/142>.
>



-- 
Byung-Gon Chun

[...]

PR #143, opened 2015-11-10
Introduce a new Matrix interface and implementation based on JBLAS

beomyeol:
Since `Nd4j` has several problem as mentioned in #142, this pull request introduces a new `Matrix` interface that supports `float` data type and will replace `Nd4j` library, and an implementation based on [jblas](http://jblas.org/).
The interface of `Matrix` is motivated by jblas's `FloatMatrix`, but does not have all methods of `FloatMatrix`. We can add methods which have not been implemented later if needed. 
In addition, this pull request also introduces `Function` interface for activation functions and `MatrixUtils`, a utility class for `Matrix`.
The implementation of a sigmoid function is included in this PR and other implementations such as ReLU, Softmax, Tanh will be added later. `MatrixUtils` provides functions for comparison and `readNumpy()` for loading matrix data from Numpy-compatible plain text files.

### DIFF
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
@@ -33,47 +33,47 @@ public interface Matrix {
   /**
    * Returns a element specified by the given index (linear indexing).
    */
-  float get(final int i);
+  float get(int i);
 
   /**
    * Returns elements specified by the linear indices.
    */
-  Matrix get(final int[] indices);
+  Matrix get(int[] indices);
 
   /**
    * Returns a element specified by the row and column indices.
    */
-  float get(final int rowIndex, final int columnIndex);
+  float get(int rowIndex, int columnIndex);
 
   /**
    * Sets a matrix element (linear indexing).
    */
-  Matrix put(final int i, final float v);
+  Matrix put(int i, float v);
 
   /**
    * Sets a matrix element.
    */
-  Matrix put(final int rowIndex, final int columnIndex, final float v);
+  Matrix put(int rowIndex, int columnIndex, float v);
 
   /**
    * Sets a column with the given column vector.
    */
-  void putColumn(final int i, Matrix v);
+  void putColumn(int i, Matrix v);
 
   /**
    * Sets a row with the given row vector.
    */
-  void putRow(final int i, Matrix v);
+  void putRow(int i, Matrix v);
 
   /**
    * Returns a copy of a column.
    */
-  Matrix getColumn(final int i);
+  Matrix getColumn(int i);
 
   /**
    * Returns a copy of a row.
    */
-  Matrix getRow(final int i);
+  Matrix getRow(int i);
 
   /**
    * Returns total number of elements.
@@ -93,13 +93,13 @@ public interface Matrix {
   /**
    * Sets all elements to the specified value.
    */
-  Matrix fill(final float v);
+  Matrix fill(float v);
 
   /**
    * Reshapes the matrix.
    * The number of elements must not change.
    */
-  Matrix reshape(final int newRows, final int newColumns);
+  Matrix reshape(int newRows, int newColumns);
 
   /**
    * Returns a string representation of this matrix.
@@ -116,7 +116,7 @@ public interface Matrix {
    * @param m a source matrix
    * @return a source matrix {@code m}
    */
-  Matrix copy(final Matrix m);
+  Matrix copy(Matrix m);
 
   /**
    * @return a duplicate of this matrix.
@@ -134,212 +134,212 @@ public interface Matrix {
   /**
    * Adds a scalar.
    */
-  Matrix add(final float v);
+  Matrix add(float v);
 
   /**
    * Adds a scalar (in place).
    */
-  Matrix addi(final float v);
+  Matrix addi(float v);
 
   /**
    * Adds a matrix.
    */
-  Matrix add(final Matrix m);
+  Matrix add(Matrix m);
 
   /**
    * Adds a matrix (in place).
    */
-  Matrix addi(final Matrix m);
+  Matrix addi(Matrix m);
 
   /**
    * Adds a vector to all columns of the matrix.
    */
-  Matrix addColumnVector(final Matrix v);
+  Matrix addColumnVector(Matrix v);
 
   /**
    * Adds a vector to all columns of the matrix (in place).
    */
-  Matrix addiColumnVector(final Matrix v);
+  Matrix addiColumnVector(Matrix v);
 
   /**
    * Adds a vector to all rows of the matrix.
    */
-  Matrix addRowVector(final Matrix v);
+  Matrix addRowVector(Matrix v);
 
   /**
    * Adds a vector to all rows of the matrix (in place).
    */
-  Matrix addiRowVector(final Matrix v);
+  Matrix addiRowVector(Matrix v);
 
   /**
    * Subtracts a scalar.
    */
-  Matrix sub(final float v);
+  Matrix sub(float v);
 
   /**
    * Subtracts a scalar (in place).
    */
-  Matrix subi(final float v);
+  Matrix subi(float v);
 
   /**
    * Subtracts a matrix.
    */
-  Matrix sub(final Matrix m);
+  Matrix sub(Matrix m);
 
   /**
    * Subtracts a matrix (in place).
    */
-  Matrix subi(final Matrix m);
+  Matrix subi(Matrix m);
 
   /**
    * Subtracts a vector to all columns of the matrix.
    */
-  Matrix subColumnVector(final Matrix v);
+  Matrix subColumnVector(Matrix v);
 
   /**
    * Subtracts a vector to all columns of the matrix (in place).
    */
-  Matrix subiColumnVector(final Matrix v);
+  Matrix subiColumnVector(Matrix v);
 
   /**
    * Subtracts a vector to all rows of the matrix.
    */
-  Matrix subRowVector(final Matrix v);
+  Matrix subRowVector(Matrix v);
 
   /**
    * Subtracts a vector to all rows of the matrix (in place).
    */
-  Matrix subiRowVector(final Matrix v);
+  Matrix subiRowVector(Matrix v);
 
   /**
    * (right-)subtracts a scalar.
    */
-  Matrix rsub(final float v);
+  Matrix rsub(float v);
 
   /**
    * (right-)subtracts a scalar (in place).
    */
-  Matrix rsubi(final float v);
+  Matrix rsubi(float v);
 
   /**
    * (right-)subtracts a matrix.
    */
-  Matrix rsub(final Matrix m);
+  Matrix rsub(Matrix m);
 
   /**
    * (right-)subtracts a matrix (in place).
    */
-  Matrix rsubi(final Matrix m);
+  Matrix rsubi(Matrix m);
 
   /**
    * Multiplies a scalar.
    */
-  Matrix mul(final float v);
+  Matrix mul(float v);
 
   /**
    * Multiplies a scalar (in place).
    */
-  Matrix muli(final float v);
+  Matrix muli(float v);
 
   /**
    * Multiplies a matrix.
    */
-  Matrix mul(final Matrix m);
+  Matrix mul(Matrix m);
 
   /**
    * Multiplies a matrix (in place).
    */
-  Matrix muli(final Matrix m);
+  Matrix muli(Matrix m);
 
   /**
    * Multiplies a vector to all columns of the matrix.
    */
-  Matrix mulColumnVector(final Matrix v);
+  Matrix mulColumnVector(Matrix v);
 
   /**
    * Multiplies a vector to all columns of the matrix (in place).
    */
-  Matrix muliColumnVector(final Matrix v);
+  Matrix muliColumnVector(Matrix v);
 
   /**
    * Multiplies a vector to all rows of the matrix.
    */
-  Matrix mulRowVector(final Matrix v);
+  Matrix mulRowVector(Matrix v);
 
   /**
    * Multiplies a vector to all rows of the matrix (in place).
    */
-  Matrix muliRowVector(final Matrix v);
+  Matrix muliRowVector(Matrix v);
 
   /**
    * Divides by a scalar.
    */
-  Matrix div(final float v);
+  Matrix div(float v);
 
   /**
    * Divides by a scalar (in place).
    */
-  Matrix divi(final float v);
+  Matrix divi(float v);
 
   /**
    * Divides by a matrix.
    */
-  Matrix div(final Matrix m);
+  Matrix div(Matrix m);
 
   /**
    * Divides by a matrix (in place).
    */
-  Matrix divi(final Matrix m);
+  Matrix divi(Matrix m);
 
   /**
    * Divides a vector to all columns of the matrix.
    */
-  Matrix divColumnVector(final Matrix v);
+  Matrix divColumnVector(Matrix v);
 
   /**
    * Divides a vector to all columns of the matrix (in place).
    */
-  Matrix diviColumnVector(final Matrix v);
+  Matrix diviColumnVector(Matrix v);
 
   /**
    * Divides a vector to all rows of the matrix.
    */
-  Matrix divRowVector(final Matrix v);
+  Matrix divRowVector(Matrix v);
 
   /**
    * Divides a vector to all rows of the matrix (in place).
    */
-  Matrix diviRowVector(final Matrix v);
+  Matrix diviRowVector(Matrix v);
 
   /**
    * (right-)divides by a scalar.
    */
-  Matrix rdiv(final float v);
+  Matrix rdiv(float v);
 
   /**
    * (right-)divides by a scalar (in place).
    */
-  Matrix rdivi(final float v);
+  Matrix rdivi(float v);
 
   /**
    * (right-)divides by a matrix.
    */
-  Matrix rdiv(final Matrix m);
+  Matrix rdiv(Matrix m);
 
   /**
    * (right-)divides by a matrix (in place).
    */
-  Matrix rdivi(final Matrix m);
+  Matrix rdivi(Matrix m);
 
   /**
    * Matrix-Matrix multiplication.
    */
-  Matrix mmul(final Matrix m);
+  Matrix mmul(Matrix m);
 
   /**
    * Matrix-Matrix multiplication (in place).
    */
-  Matrix mmuli(final Matrix m);
+  Matrix mmuli(Matrix m);
 
   /**
    * Returns the maximum element of the matrix.
@@ -392,5 +392,5 @@ public interface Matrix {
    * Returns true if and only if the given matrix has the same size
    * and the maximal absolute difference in all elements is smaller than the specified tolerance.
    */
-  boolean compare(final Matrix m, float tolerance);
+  boolean compare(Matrix m, float tolerance);
 }
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
@@ -342,6 +342,51 @@ public interface Matrix {
   Matrix mmuli(final Matrix m);
 
   /**
+   * Returns the maximum element of the matrix.
+   */
+  float max();
+
+  /**
+   * Returns column-wise maximums.
+   */
+  Matrix columnMaxs();
+
+  /**
+   * Returns row-wise maximums.
+   */
+  Matrix rowMaxs();
+
+  /**
+   * Returns the minimum element of the matrix.
+   */
+  float min();
+
+  /**
+   * Returns column-wise minimums.
+   */
+  Matrix columnMins();
+
+  /**
+   * Returns row-wise minimums.
+   */
+  Matrix rowMins();
+
+  /**
+   * Returns a row vector containing the sum of elements in each column.
+   */
+  Matrix columnSums();
+
+  /**
+   * Returns a column vector containing the sum of elements in each row.
+   */
+  Matrix rowSums();
+
+  /**
+   * Returns the sum of all elements in the matrix.
+   */
+  float sum();
+
+  /**
    * Compares with the given matrix.
    *
    * Returns true if and only if the given matrix has the same size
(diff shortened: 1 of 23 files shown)


beomyeol:
I've added ReLU, identity, absolute value, square and hyperbolic tangent function. Please take a look at these too.

beomyeol:
I changed the assumption that the argument of `derivative()` method is the return value of `apply()`. In this case, derivates of the absolute value function and the square function cannot be derived from their return values because the signs of original inputs are lost by `apply()` operation. Thus, I removed the absolute value and square functions.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
```diff
@@ -0,0 +1,396 @@
...
+   */
+  int getRows();
+
+  /**
+   * Returns the number of columns.
+   */
+  int getColumns();
+
+  /**
+   * Returns a element specified by the given index (linear indexing).
```
jsjason: Please include comments on how this 'linear indexing' is processed: row-wise or column-wise?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
```diff
@@ -0,0 +1,396 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package edu.snu.dolphin.dnn.blas;
+
+/**
+ * Matrix interface.
+ */
+public interface Matrix {
```
jsjason: This interface's method comments are a bit confusing; some `get` methods return 'copies', while some don't. Is this intentional?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
```diff
@@ -0,0 +1,396 @@
...
+   */
+  Matrix get(final int[] indices);
+
+  /**
+   * Returns a element specified by the row and column indices.
+   */
+  float get(final int rowIndex, final int columnIndex);
+
+  /**
+   * Sets a matrix element (linear indexing).
```
jsjason: Please include comments on how this 'linear indexing' is processed: row-wise or column-wise?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
```diff
@@ -0,0 +1,396 @@
...
+
+  /**
+   * Returns a element specified by the given index (linear indexing).
+   */
+  float get(final int i);
+
+  /**
+   * Returns elements specified by the linear indices.
+   */
+  Matrix get(final int[] indices);
```
jsjason: This method should have more explanation on the returned matrix's shape.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
```diff
@@ -0,0 +1,396 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package edu.snu.dolphin.dnn.blas;
+
+/**
+ * Matrix interface.
```
jsjason: Please put more comments explaining that this matrix's elements are floats.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
```diff
@@ -0,0 +1,396 @@
...
+
+  /**
+   * Returns a element specified by the row and column indices.
+   */
+  float get(final int rowIndex, final int columnIndex);
+
+  /**
+   * Sets a matrix element (linear indexing).
+   */
+  Matrix put(final int i, final float v);
```
jsjason: The variable name `i` is inconsistent with `put(rowIndex, columnIndex, v)`. We should use `index` or `linearIndex`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
```diff
@@ -0,0 +1,396 @@
...
+
+  /**
+   * Converts the matrix to a one-dimensional array of {@code float}s.
+   */
+  float[] toFloatArray();
+
+  /**
+   * Copies {@code Matrix} {@code m} to this.
+   * @param m a source matrix
+   * @return a source matrix {@code m}
```
jsjason: Why would we want to return the argument as it is if it's not modified?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
```diff
@@ -0,0 +1,396 @@
...
+
+  /**
+   * Returns the number of columns.
+   */
+  int getColumns();
+
+  /**
+   * Returns a element specified by the given index (linear indexing).
+   */
+  float get(final int i);
```
jsjason: Putting `final` on interface method arguments isn't REEF/Dolphin's coding style.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
```diff
@@ -0,0 +1,396 @@
...
+  /**
+   * Returns transposed copy of this matrix.
+   */
+  Matrix transpose();
+
+
+  /* Operations */
+
+  /**
+   * Adds a scalar.
```
jsjason: Could we change the this line to `Adds a scalar to all elements.`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
```diff
@@ -0,0 +1,396 @@
...
+   */
+  Matrix mul(final float v);
+
+  /**
+   * Multiplies a scalar (in place).
+   */
+  Matrix muli(final float v);
+
+  /**
+   * Multiplies a matrix.
```
jsjason: Could we include the term `element-wise`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
```diff
@@ -0,0 +1,396 @@
...
+   */
+  Matrix mul(final Matrix m);
+
+  /**
+   * Multiplies a matrix (in place).
+   */
+  Matrix muli(final Matrix m);
+
+  /**
+   * Multiplies a vector to all columns of the matrix.
```
jsjason: I'm not sure what this means. Does "multiply a vector to a column" mean calculating the dot product? Please add comments.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixFactory.java
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * Factory interface for {@link Matrix}.
+ */
+public interface MatrixFactory {
+
+  /**
+   * Creates a row vector.
+   * @param length the length of a row vector
+   * @return a generated row vector
```
jsjason: How is this vector initialized?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixFactory.java
```diff
@@ -0,0 +1,148 @@
...
+  /**
+   * Creates a row vector with the given values.
+   * @param data elements of a row vector
+   * @return a generated row vector
+   */
+  Matrix create(final float[] data);
+
+  /**
+   * Creates a matrix with the given values.
+   * @param data elements of a matrix
```
jsjason: Which dimension corresponds to rows/columns? Please add comments.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixFactory.java
```diff
@@ -0,0 +1,148 @@
...
+   * Creates a matrix with random values uniformly distributed in 0..1.
+   * @param rows the number of rows
+   * @param columns the number of columns
+   * @param seed a random seed
+   * @return a generated matrix
+   */
+  Matrix rand(final int rows, final int columns, long seed);
+
+  /**
+   * Creates a row vector with normally distributed random values.
```
jsjason: What's the mean and variance of this distribution?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixFunctions.java
```diff
@@ -0,0 +1,91 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package edu.snu.dolphin.dnn.blas;
+
+/**
+ * Utility class that provides functions for {@link edu.snu.dolphin.dnn.blas.Matrix}.
```
jsjason: The link doesn't need to include package names in case the packages are the same.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixUtils.java
```diff
@@ -0,0 +1,136 @@
...
+package edu.snu.dolphin.dnn.blas;
+
+import java.io.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Utility class for {@link edu.snu.dolphin.dnn.blas.Matrix}.
```
jsjason: The link doesn't need to include package names in case the packages are the same.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixUtils.java
```diff
@@ -0,0 +1,136 @@
...
+   * @param tolerance the maximum difference for which both numbers are still considered equal.
+   * @return true if the two specified arrays of Matrix are equal to one another.
+   */
+  public static boolean compare(final Matrix[] a, final Matrix[] b, final float tolerance) {
+    return compare(Arrays.asList(a), Arrays.asList(b), tolerance);
+  }
+
+
+  /**
+   * Loads a matrix from a Numpy-compatible plain text file with the specified delimiter.
```
jsjason: The comments say "text file", but the actual parameter is an input stream.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixUtils.java
```diff
@@ -0,0 +1,136 @@
...
+   * @param delimiter a delimiter
+   * @return a loaded matrix
+   * @throws IOException
+   */
+  public static Matrix readNumpy(final MatrixFactory matrixFactory,
+                                 final InputStream inputStream,
+                                 final String delimiter) throws IOException {
+    final BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream));
+    String line;
+    final List<float[]> data2 = new ArrayList<>();
```
jsjason: I didn't notice it last time, but the variable name `data2` seems inappropriate. Could we use a better name?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixUtils.java
```diff
@@ -0,0 +1,136 @@
...
+    int numColumns = -1;
+    final Matrix ret;
+    while ((line = reader.readLine()) != null) {
+      final String[] data = line.trim().split(delimiter);
+      if (numColumns < 0) {
+        numColumns = data.length;
+
+      } else {
+        assert data.length == numColumns : "Data has inconsistent number of columns";
```
jsjason: Have you checked that assertion is enabled on REEF jobs?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/function/Identity.java
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+package edu.snu.dolphin.dnn.blas.function;
+
+import edu.snu.dolphin.dnn.blas.Matrix;
+
+/**
+ * Identity function.
+ */
+public final class Identity implements Function {
```
jsjason: Can we remove `public`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
```diff
@@ -0,0 +1,396 @@
...
+
+  /**
+   * Copies {@code Matrix} {@code m} to this.
+   * @param m a source matrix
+   * @return a source matrix {@code m}
+   */
+  Matrix copy(final Matrix m);
+
+  /**
+   * @return a duplicate of this matrix.
```
jsjason: This method and the one above are the only ones that use `@return`. Let's use a consistent style.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixUtils.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  /**
+   * Creates a row vector in which only element at the specified position is {@code 1} and other elements are {@code 0}.
+   *
+   * @param matrixFactory a matrix factor used to create a matrix
+   * @param index an index of the element to be set to {@code}
+   * @param length the length of a row vector
+   * @return a generated row vector
+   * @return
```
jsjason: Redunant line.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/function/Function.java
```diff
@@ -0,0 +1,53 @@
...
+
+  Matrix apply(final Matrix m);
+
+  Matrix applyi(final Matrix m);
+
+  Matrix derivative(final Matrix m);
+
+  Matrix derivativei(final Matrix m);
+
+  final class Factory {
```
jsjason: Why do we declare this class an inner class of `Function`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/function/Function.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private Factory() {
+    }
+
+    public static Function getFunction(final String name) {
+      switch (name.toLowerCase()) {
+      case "identity":
+        return new Identity();
```
jsjason: Do we instantiate a new class every time `getFunction` is called?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/function/ReLU.java
```diff
@@ -0,0 +1,68 @@
...
+ * limitations under the License.
+ */
+package edu.snu.dolphin.dnn.blas.function;
+
+import edu.snu.dolphin.dnn.blas.Matrix;
+
+/**
+ * Rectified linear unit.
+ */
+public final class ReLU implements Function {
```
jsjason: Can we remove `public`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/function/Tanh.java
```diff
@@ -0,0 +1,61 @@
...
+   */
+  @Override
+  public Matrix derivative(final Matrix m) {
+    return derivativei(m.dup());
+  }
+
+  /**
+   * Calculates the matrix in which all elements are derivatives of a hyperbolic tangent function (in place).
+   * <p>
+   *   derivatives of tanh: 1 - tanh(x)^2
```
jsjason: typo: "derivative of tanh"

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/function/Identity.java
```diff
@@ -0,0 +1,57 @@
...
+
+  /**
+   * Applies an identity function to all elements of the specified matrix.
```
jsjason: "Applies ~~an~~**the** identity function to..."
Same for all other comments.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/function/Tanh.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+package edu.snu.dolphin.dnn.blas.function;
+
+import edu.snu.dolphin.dnn.blas.Matrix;
+import edu.snu.dolphin.dnn.blas.MatrixFunctions;
+
+/**
+ * Hyperbolic tangent function.
+ */
+public final class Tanh implements Function {
```
jsjason: Can we remove `public`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/jblas/MatrixJBLASImpl.java
```diff
@@ -0,0 +1,485 @@
...
+    }
+    return false;
+  }
+
+  @Override
+  public int hashCode() {
+    return matrix.hashCode();
+  }
+
+  public static MatrixJBLASImpl concatHorizontally(final MatrixJBLASImpl a, final MatrixJBLASImpl b) {
```
jsjason: According to the current implementation, to concat two `Matrix`s, we need to receive an injection of a `MatrixFactory` and use `MatrixFactory.concatxx()`. Couldn't we just have concatenation as a method of the `Matrix` implementation?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/jblas/MatrixJBLASImpl.java
```diff
@@ -0,0 +1,485 @@
...
+ */
+package edu.snu.dolphin.dnn.blas.jblas;
+
+import edu.snu.dolphin.dnn.blas.Matrix;
+import org.jblas.FloatMatrix;
+
+/**
+ * Matrix implementation based on JBLAS.
+ */
+public class MatrixJBLASImpl implements Matrix {
```
jsjason: Is it correct to make this class `public`?

jsjason:
I did a pass, @beomyeol. It would've been painful to write this code; thanks a lot!

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixFactory.java
beomyeol: For the matrix implementation based on 'jblas', the returned vector is initialized with zeros by [Java spec](http://docs.oracle.com/javase/specs/jls/se7/html/jls-4.html#jls-4.12.5). So, `create()` and `zeros()` make the same results. However, it is possible that the returned vector may not be initialized depending on the backend libraries.
Do we should change the comment to "Creates an uninitialized row vector"?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
beomyeol: I followed the interface of jblas's one. But, returning `this` seems better since many other methods return `this`. I'll change this.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
beomyeol: I'll change this to "element-wise multiplication". Please let me know if this is still confusing.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/Matrix.java
beomyeol: @jsjason and I discussed this. One thing that may be confusing is `get()` with `int[]` as the argument. This method returns a new row vector and it is not a copy of whole or part of the target matrix. But, to make it less confusing, I'll add a comment that the modification on the returned vector does not affect the original matrix

beomyeol:
@jsjason I addressed your comments. Please take another look.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixFactory.java
jsjason: No, it looks fine.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/jblas/MatrixJBLASFactory.java
```diff
@@ -0,0 +1,167 @@
...
+
+    if (a.getColumns() != b.getColumns()) {
+      throw new RuntimeException("Matrices do not have the same number of columns");
+    } else {
+      final Matrix ret = create(a.getRows() + b.getRows(), a.getColumns());
+      for (int i = 0; i < a.getRows(); ++i) {
+        ret.putRow(i, a.getRow(i));
+      }
+      for (int i = 0; i < b.getRows(); ++i) {
+        ret.putRow(a.getColumns() + i, b.getRow(i));
```
jsjason: Is this the right implementation?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/jblas/MatrixJBLASImpl.java
```diff
@@ -0,0 +1,488 @@
...
+ */
+package edu.snu.dolphin.dnn.blas.jblas;
+
+import edu.snu.dolphin.dnn.blas.Matrix;
+import org.jblas.FloatMatrix;
+
+/**
+ * Matrix implementation based on JBLAS.
+ */
+class MatrixJBLASImpl implements Matrix {
```
jsjason: Is this not made final on purpose?

jsjason:
@beomyeol This looks great. I'll merge this after you address my new comments.

beomyeol:
@jsjason Thanks for your review. I've addressed your new comments :)

PR #145, opened 2015-11-15
Separates activation functions from fully connected layers and Introduces activation layers

### LINKED ISSUE #144: Separate activation function from fully connected layer
Now, activation function is a part of `FullyConnectedLayer`. However, in many neural network models such as [AlexNet](https://github.com/BVLC/caffe/tree/master/models/bvlc_alexnet) and [GoogleNet](https://github.com/BVLC/caffe/tree/master/models/bvlc_googlenet), activation function `ReLU` is used for the outputs of convolutional layers and fully connected layers.
We can add the activation function feature to other layers. Nonetheless, in order to reduce duplicates and be more general, it is better to have separate layers for activation functions like Caffe and Apache SINGA.

beomyeol:
This pull request removes activation functions from fully connected layers and introduces activation layers. The activation layer applies the activation function specified by the configuration file to its input value.

This pull request includes the following changes
* Changes API of `LayerBase`'s '`backPropagate()` method.
* Introduces `DefaultLayerParameterInitializer` and an empty layer parameter `EMPTY` for layers which are not learnable.
* Changes the type of exceptions related to invalid indices from `RuntimeException` to `IllegalArgumentException`.
* Fixes bugs related to indices in `backPropagateFromTo()`.
* Fixes typo `isForward`.
* Adds test for activation layers.

The change of `LayerBase`'s `backPropagate()` method API includes the removal of `nextParameter` and the introduction of `input`. The introduction of activation layers makes `nextParameter` unnecessary in `backPropagate()`. Also, although the input value is not needed now, it will be needed for computing gradients of some functions such as absolute value and power functions.

This closes #144.

### DIFF
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetwork.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetwork.java
@@ -169,20 +169,25 @@ public final class NeuralNetwork {
   }
 
   /**
-   * Computes errors from output layer to input layer.
+   * Computes errors from the output layer to the input layer.
+   * This returns an empty array when the network model has zero or one layer.
    * @param activations an array of activations for each layer.
    * @param label the expected output.
    * @return an array of errors for each layer.
    */
   public INDArray[] backPropagate(final INDArray[] activations,
                                   final INDArray label) {
-    // Process backpropagation to the second layer
-    // because the error the first layer generates, is not needed to generate a gradient for learning the first layer.
+    // Process backpropagation to the second layer.
+    // because the error returned by the first layer's backpropagation, is not needed
+    // to generate gradients for learning the first layer.
+    // The errors for generating gradients used to update the first layer's parameter, are calculated
+    // in the next layer's backpropagation.
     return backPropagateTo(1, activations, label);
   }
 
   /**
-   * Computes errors from output layer to the specified ending layer.
+   * Computes errors from the output layer to the specified ending layer.
+   * This returns an empty array when the network model has zero or one layer.
    * @param end the index of ending layer, inclusive.
    * @param activations an array of activations of each layer.
    * @param label the expected output.
@@ -191,16 +196,25 @@ public final class NeuralNetwork {
   public INDArray[] backPropagateTo(final int end,
                                     final INDArray[] activations,
                                     final INDArray label) {
-    final int lastLayerIndex = layers.length - 1;
-
-    // The first element of activations is the input data.
-    // So, (i + 1)-th element of activations refers to the activation of i-th layer.
-    final INDArray error = layers[lastLayerIndex].backPropagate(label, activations[lastLayerIndex + 1], EMPTY);
-
-    if (lastLayerIndex == end) {
-      return new INDArray[]{error};
+    // Case 1: Only one layer
+    if (layers.length < 2) {
+      // If a neural network has only one layer, the network does not process backpropagation for this layer because
+      // this layer cannot generate gradients for updating its parameter.
+      // Generating gradients requires the error computed by the next layer.
+      return new INDArray[0];
     } else {
-      return ArrayUtils.add(backPropagateFromTo(lastLayerIndex - 1, end, activations, error), error);
+      final int lastLayerIndex = layers.length - 1;
+      // The first element of activations is the input data.
+      // So, (i + 1)-th element of activations refers to the activation of i-th layer.
+      final INDArray error = layers[lastLayerIndex].backPropagate(label, activations[lastLayerIndex + 1], EMPTY);
+
+      // Case 2: Two layers
+      if (lastLayerIndex == end) {
+        return new INDArray[]{error};
+      // Case 3: More than two layers
+      } else {
+        return ArrayUtils.add(backPropagateFromTo(lastLayerIndex - 1, end, activations, error), error);
+      }
     }
   }
 
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetwork.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetwork.java
@@ -172,6 +172,8 @@ public final class NeuralNetwork {
    */
   public INDArray[] backPropagate(final INDArray[] activations,
                                   final INDArray label) {
+    // Process backpropagation to the second layer
+    // because the error the first layer generates, is not needed to generate a gradient for learning the first layer.
     return backPropagateTo(1, activations, label);
   }
 
@@ -185,11 +187,6 @@ public final class NeuralNetwork {
   public INDArray[] backPropagateTo(final int end,
                                     final INDArray[] activations,
                                     final INDArray label) {
-    if (end == 0) {
-      throw new IllegalArgumentException("The ending layer cannot be the first layer: " +
-          "the error of an input value does not make sense.");
-    }
-
     final int lastLayerIndex = layers.length - 1;
 
     // The first element of activations is the input data.
@@ -211,6 +208,11 @@ public final class NeuralNetwork {
                                         final INDArray nextError) {
     if (begin == layers.length - 1) {
       throw new IllegalArgumentException("The beginning layer of backPropagateFromTo cannot be the output layer");
+    }
+    if (end == 0) {
+      // The errors for generating gradients of the first layer are calculated by the next layer, not the first layer.
+      throw new IllegalArgumentException("The ending layer cannot be the first layer: " +
+          "The error that is propagated to the input is unnecessary to generate gradients for the first layer");
     }
     if (begin < end) {
       throw new IllegalArgumentException(String.format(
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetwork.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetwork.java
@@ -147,7 +147,7 @@ public final class NeuralNetwork {
    */
   public INDArray[] feedForward(final int begin, final int end, final INDArray input) {
     if (begin > end) {
-      throw new RuntimeException(String.format(
+      throw new IllegalArgumentException(String.format(
           "The beginning index (%d) must be less than or equal to the ending index (%d).", begin, end));
     }
 
@@ -172,7 +172,7 @@ public final class NeuralNetwork {
    */
   public INDArray[] backPropagate(final INDArray[] activations,
                                   final INDArray label) {
-    return backPropagateTo(0, activations, label);
+    return backPropagateTo(1, activations, label);
   }
 
   /**
@@ -185,9 +185,14 @@ public final class NeuralNetwork {
   public INDArray[] backPropagateTo(final int end,
                                     final INDArray[] activations,
                                     final INDArray label) {
+    if (end == 0) {
+      throw new IllegalArgumentException("The ending layer cannot be the first layer: " +
+          "the error of an input value does not make sense.");
+    }
+
     final int lastLayerIndex = layers.length - 1;
 
-    // The first element of activations is input data.
+    // The first element of activations is the input data.
     // So, (i + 1)-th element of activations refers to the activation of i-th layer.
     final INDArray error = layers[lastLayerIndex].backPropagate(activations[lastLayerIndex + 1], label);
     return ArrayUtils.add(backPropagateFromTo(lastLayerIndex - 1, end, activations, error), error);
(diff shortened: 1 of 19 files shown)


dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetwork.java
```diff
@@ -185,9 +185,14 @@ public void train(final INDArray input, final int label) {
   public INDArray[] backPropagateTo(final int end,
                                     final INDArray[] activations,
                                     final INDArray label) {
+    if (end == 0) {
+      throw new IllegalArgumentException("The ending layer cannot be the first layer: " +
```
jsjason: I'm not sure if this is correct. We don't have a separate 'input layer', do we? For example, the first layer in the sample conf file is a fully connected layer, which can have errors.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetwork.java
```diff
@@ -205,29 +210,29 @@ public void train(final INDArray input, final int label) {
...
+    checkIndices(begin, end, false);
 
-    final INDArray[] errors = new INDArray[end - begin + 1];
+    final INDArray[] errors = new INDArray[begin - end + 1];
     INDArray error = nextError;
 
     for (int i = begin; i >= end; --i) {
-      error = layers[i].backPropagate(activations[i + 1], layers[i + 1].getLayerParameter(), error);
-      errors[i - begin] = error;
+      error = layers[i].backPropagate(activations[i], activations[i + 1], error);
```
jsjason: This changes which layer we're calling `getLayerParameter` on. Originally, we were using the *next* layer's parameters to compute the weight, but now we're using *this* layer's parameters. Could you explain in more detail what was changed?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/ActivationLayerConfigurationBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  @Override
+  public synchronized Configuration build() {
+    return Tang.Factory.getTang().newConfigurationBuilder()
+        .bindNamedParameter(LayerConfigurationParameters.NumberOfInput.class, String.valueOf(numInput))
+        .bindNamedParameter(LayerConfigurationParameters.NumberOfOutput.class, String.valueOf(numOutput))
+        .bindNamedParameter(LayerConfigurationParameters.ActivationFunction.class, String.valueOf(activationFunction))
+        .bindImplementation(LayerBase.class, ActivationLayer.class)
+        .bindImplementation(LayerParameterInitializer.class, DefaultLayerParameterInitializer.class)
```
jsjason: Can we make `DefaultLayerParameterInitializer` as the default implementation of `LayerParameterInitializer` and remove this line?

dolphin-dnn/src/test/java/edu/snu/dolphin/dnn/NeuralNetworkTest.java
```diff
@@ -119,8 +137,8 @@ public void buildNeuralNetwork() throws InjectionException {
   @Test
   public void feedForwardTest() {
     final INDArray[] activations = neuralNetwork.feedForward(input);
-    assertTrue(Nd4jUtils.equals(activations, expectedActivations, tolerance));
     assertTrue(Nd4jUtils.equals(activations[activations.length - 1], expectedOutput, tolerance));
+    assertTrue(Nd4jUtils.equals(activations, expectedActivations, tolerance));
```
jsjason: What's this change for?

jsjason:
@beomyeol The tests succeed but the example finished with an error (although, the training did progress according to the logs).
```
2015-11-19 15:36:56,171 INFO edu.snu.dolphin.dnn.NeuralNetworkREEF.main main | REEF job completed: FAILED(java.lang.NullPointerException: Null ids parameter.)
```
Am I the only one that met this error?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetwork.java
beomyeol: The error that the first layer calculates is not used for generating gradients for the first layer. Even though the first layer is a fully connected layer, the errors needed for generating gradients are calculated by the next layer's backpropagation. Thus, the backpropagation for the first layer is unnecessary.
I'll add more comments to this.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetwork.java
beomyeol: In original implementation, A learnable layer (e.g. fully connected layer) calculates the errors that are used to generate gradients for training itself. For this computation, the next layer's parameter (weight and bias) is needed. 
This pull request changes that the errors used for generating gradients are calculated by the next layer. This makes the next layer's parameter unnecessary and learnable layers use their own parameters in backpropagation.

dolphin-dnn/src/test/java/edu/snu/dolphin/dnn/NeuralNetworkTest.java
beomyeol: An array of activations includes the output of a neural network. If the array of activations that the neural network generates is equal to `expectedActivations` within `tolerance`, the next line before change, that checks the equality between the last element of `activations` and `expectedOutput` was totally unnecessary.
However, if the order is reversed, we may easily figure out the cause of test failure cases such as one of activation values is not returned despite the correct output value.

beomyeol:
@jsjason I addressed your comments and left some comments about your questions.
Also, I fixed a bug that the backpropagation does not work well in a neural network with 2 layers and introduced `ActivationWithLossLayer`.

`ActivationWithLossLayer` is a wrapper of `ActivationLayer`. Its forward pass is same to that of `ActivationLayer`, but a backward pass is calculating a derivative of the configured loss function with respect to the output value and the expected output value of the layer. The cross-entropy loss function is included in this pull request. We can add other loss functions later.

The `ActivationWithLossLayer` is needed for supporting other loss functions, and for the consistency of which layer (itself or **the next layer**) calculates errors to generate gradients for training the layer.

beomyeol:
@jsjason I did not experience the error that you said. Please send me the logs when the error occurred.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetwork.java
```diff
@@ -172,7 +176,9 @@ public void train(final INDArray input, final int label) {
    */
   public INDArray[] backPropagate(final INDArray[] activations,
                                   final INDArray label) {
-    return backPropagateTo(0, activations, label);
+    // Process backpropagation to the second layer
+    // because the error the first layer generates, is not needed to generate a gradient for learning the first layer.
```
jsjason: I'm still not satisfied with this explanation. This makes it sound like the first layer is always not learnable. Also, the words "the error the first layer generates" can be misunderstood if the reader doesn't know that weight updates are done by the next layer's back propagation call.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetwork.java
```diff
@@ -187,10 +193,15 @@ public void train(final INDArray input, final int label) {
...
     final int lastLayerIndex = layers.length - 1;
 
-    // The first element of activations is input data.
+    // The first element of activations is the input data.
     // So, (i + 1)-th element of activations refers to the activation of i-th layer.
-    final INDArray error = layers[lastLayerIndex].backPropagate(activations[lastLayerIndex + 1], label);
-    return ArrayUtils.add(backPropagateFromTo(lastLayerIndex - 1, end, activations, error), error);
+    final INDArray error = layers[lastLayerIndex].backPropagate(label, activations[lastLayerIndex + 1], EMPTY);
+
+    if (lastLayerIndex == end) {
```
jsjason: Is this for cases where the activationWithLoss layer is the only layer?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/ActivationWithLossLayer.java
```diff
@@ -0,0 +1,94 @@
...
+   * Compute the error for the specified loss function.
+   * @param label the label value.
+   * @param activation the activation value.
+   * @param nextError an error of the next layer - this argument is ignored.
+   * @return the error with respect to the activation and label values.
+   */
+  @Override
+  public INDArray backPropagate(final INDArray label, final INDArray activation, final INDArray nextError) {
+    switch (lossFunction.toLowerCase()) {
+    case "cross-entropy":
```
jsjason: The other names don't use hyphens; are we sure we want to use this naming convention just for loss functions?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/ActivationWithLossLayerConfigurationBuilder.java
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.reef.tang.Configuration;
+import org.apache.reef.tang.Tang;
+import org.apache.reef.tang.formats.AvroConfigurationSerializer;
+import org.apache.reef.tang.formats.ConfigurationSerializer;
+import org.apache.reef.util.Builder;
+
+/**
+ * Configuration builder for activation with loss layer.
+ *
+ * The configuration that this builder generates is used to create an activation with layer instance.
```
jsjason: "activation with **loss** layer instance"?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/ActivationWithLossLayer.java
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.reef.tang.annotations.Parameter;
+import org.apache.reef.tang.exceptions.InjectionException;
+import org.apache.reef.tang.formats.ConfigurationSerializer;
+import org.nd4j.linalg.api.ndarray.INDArray;
+
+import javax.inject.Inject;
+import java.io.IOException;
+
+/**
+ * Loss Layer with activation function.
```
jsjason: We should have comments saying that this layer is intended to be the final layer, in most cases.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/ActivationLayer.java
```diff
@@ -0,0 +1,78 @@
...
+import edu.snu.dolphin.dnn.conf.LayerConfigurationParameters.LayerIndex;
+import edu.snu.dolphin.dnn.conf.LayerConfigurationParameters.NumberOfOutput;
+import org.apache.reef.tang.annotations.Parameter;
+import org.nd4j.linalg.api.ndarray.INDArray;
+import org.nd4j.linalg.factory.Nd4j;
+
+import javax.inject.Inject;
+
+/**
+ * Activation Layer.
```
jsjason: Shouldn't we include what this layer is actually doing, and various characteristics such as the number of inputs and outputs being the same?

jsjason:
The new ActivationWithLoss layer seems good. @beomyeol Please check my comments. I'll show you the logs later, offline.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetwork.java
beomyeol: This is the case that a neural network model has two layers and the last one is the `ActivationWithLossLayer`. I missed the case where the network model has only one layer.
In this case, the model cannot be trained since the gradients used for updating the layer should be calculated in the next layer's backpropagation. But, the next layer does not exist in the network.
I'll add a code for handling the case where the model has only one layer and comments about this.

beomyeol:
@jsjason Thanks for your review. I've addressed your comments. Please take another look.

jsjason:
@beomyeol The changes look great, but I have an extra question.
If I train a network with two FullyConnectedLayers and no ActivationWithLoss layers, I get an exception. But what I expected was that the application finishes correctly, although without any meaningful results. Is throwing an exception intended?

jsjason:
Here is the code I ran:
```Java
final Configuration nnConf = NeuralNetworkConfigurationBuilder.newConfigurationBuilder()
        .setBatchSize(1)
        .setStepsize(1e-2f)
        .setParameterProviderClass(LocalNeuralNetParameterProvider.class)
        .addLayerConfiguration(
            FullyConnectedLayerConfigurationBuilder.newConfigurationBuilder()
                .setNumInput(3)
                .setNumOutput(4)
                .setInitWeight(0.0001f)
                .setInitBias(0.0002f)
                .setRandomSeed(10)
                .build())
        .addLayerConfiguration(
            FullyConnectedLayerConfigurationBuilder.newConfigurationBuilder()
                .setNumInput(4)
                .setNumOutput(3)
                .setInitWeight(0.0001f)
                .setInitBias(0.0002f)
                .setRandomSeed(10)
                .build())
        .build();

    final Injector injector = Tang.Factory.getTang().newInjector(nnConf);
    neuralNetwork = injector.getInstance(NeuralNetwork.class);
    neuralNetwork.train(Nd4j.create(new float[]{1, 2, 3}), 1);
```

beomyeol:
@jsjason If the last layer is a fully connected layer, the backpropagation cannot work well since the fully connected layer needs the next layer's error. Because the error value that the last layer gets as the argument of the backpropagation cannot be defined. the neural network model uses the empty array as the argument of the last layer's backpropagation, which causes the exception. Thus, the last layer's backpropagation should generate an error without the next layer's error (e.g. `ActivationWithLossLayer`'s one).

jsjason:
@beomyeol Thanks. Another question: I know I'm asking you the same thing over and over again, but I'm confused yet again about excluding the first layer from back propagation. If we had n layers (not counting the input layer), shouldn't we have exactly n error vectors (matrices)? This code seems to output n-1 error vectors.

beomyeol:
@jsjason Yes, this code generate n-1 error vectors. The point that you may miss is that the last layer cannot be learnable. Let me suppose all layers except the last one are learnable. The network model can generate the parameter updates for these n-1 layers with n-1 error vectors. To be specific, the parameter updates for (n-1)-th layer is calculated with the error generated by n-th layer's backpropagation, the parameter update (n-2)-th layer is calculated with the error generated by (n-1)-th layer's one. That's why we can work well with n-1 error vectors. Please let me know if you are still confused.

jsjason:
@beomyeol Thanks, I understand what's going on in this PR now. I'll merge this after testing.

PR #148, opened 2015-12-01
Replace ND4J library with the new jblas-based matrix library

### LINKED ISSUE #142: Incorrect results of matrix multiplication and matrix concatenation of Nd4j library

beomyeol:
This pull request replaces ND4J library with the new jblas-based matrix library.
Here are the changes included in this PR.
* Replace ND4J library with the new jblas-based matrix library.
* Upgrade to the snapshot of REEF 0.14. (the snapshot should include Apache REEF's [#632](https://github.com/apache/incubator-reef/pull/632) to work well with the parameter server or group communication.)
* Add the square function, the absolute function and the softmax function.
* Change the argument of `Function`'s `derivative()` method as the value before the activation functions are applied.
* Change the activation functions of the neural network configuration example to ReLU and softmax.
* Remove `inputShape` and related codes.

The changes of this pull request make our DNN implementation lose the support of arbitrary dimensional inputs. I'll make another PR to restore this feature later.

This closes #142


jsjason:
The tests have failed with
```
java.lang.UnsatisfiedLinkError: org.jblas.NativeBlas.sgemm(CCIIIF[FII[FIIF[FII)V
```
What's the problem?

jsjason:
We needed to install libgfortran3 on the CI server to resolve the test failures.

beomyeol:
For 64-bit Linux and Windows, additional libraries need to be installed for using jblas. https://github.com/mikiobraun/jblas/wiki/Missing-Libraries

README.md
```diff
@@ -10,7 +10,11 @@ Dolphin is a machine learning platform built on top of [Apache REEF](http://reef
 * [`dolphin-ps`](dolphin-ps/README.md): Parameter server module for asynchronous machine learning algorithms.
 
 ### How to build and run Dolphin
-1. Build Dolphin:
+
+1. Build REEF: check https://cwiki.apache.org/confluence/display/REEF/Compiling+REEF
+  Currently, Dolphin depends on REEF `0.14.0-incubating-SNAPSHOT`, which means you must build the current snapshot of REEF before building Dolphin. We will move to the release version `0.14.0-incubating` once it's out.
```
jsjason: Actually, the current snapshot version of REEF is `0.14.0-SNAPSHOT`; no more incubating!

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetworkParameterUpdater.java
```diff
@@ -231,12 +238,15 @@ public NeuralNetParamServerData initValue(final String key) {
...
+    final Configuration matrixFactoryConf = Tang.Factory.getTang().newConfigurationBuilder()
+        .bindImplementation(MatrixFactory.class, matrixFactory.getClass())
+        .build();
 
     for (final String serializedInitializerConfiguration : serializedLayerConfigurationSet) {
       try {
         final Configuration initializerConfiguration =
             configurationSerializer.fromString(serializedInitializerConfiguration);
-        final Injector injector = Tang.Factory.getTang().newInjector(initializerConfiguration);
+        final Injector injector = Tang.Factory.getTang().newInjector(initializerConfiguration, matrixFactoryConf);
```
jsjason: Can we receive an Injector at the constructor and call `injector.forkInjector()` here? That way, we don't have to create `matrixFactoryConf`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetwork.java
```diff
@@ -61,22 +63,38 @@
...
   private int trainedCount;
 
-  private static final INDArray EMPTY = Nd4j.create(0);
-
+  /**
+   * the empty matrix.
+   * This is used as the next error of the last layer's backpropagation.
+   */
+  private final Matrix emptyMatrix;
+  /**
```
jsjason: We should put a blank line here.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/NeuralNetworkConfigurationBuilder.java
```diff
@@ -63,8 +64,8 @@ public synchronized NeuralNetworkConfigurationBuilder setBatchSize(final int bat
   }
 
   public synchronized NeuralNetworkConfigurationBuilder setParameterProviderClass(
-      final Class<? extends ParameterProvider> parameterProviderClass) {
-    this.parameterProviderClass = parameterProviderClass;
+      final Class<? extends ParameterProvider> newParameterProviderClass) {
```
jsjason: What is this change for?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/GroupCommParameterServerTask.java
```diff
@@ -79,17 +81,23 @@ private GroupCommParameterServerTask(
...
+        .bindImplementation(MatrixFactory.class, matrixFactory.getClass())
+        .build();
+
     for (final String serializedInitializerConfiguration : serializedLayerConfigurationSet) {
       try {
         final Configuration initializerConfiguration =
             configurationSerializer.fromString(serializedInitializerConfiguration);
         final LayerParameterInitializer layerParameterInitializer =
-            Tang.Factory.getTang().newInjector(initializerConfiguration).getInstance(LayerParameterInitializer.class);
+            Tang.Factory.getTang().newInjector(initializerConfiguration, matrixFactoryConf)
```
jsjason: Can we receive an Injector at the constructor and call `injector.forkInjector()` here? That way, we don't have to create `matrixFactoryConf`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetworkGroupCommDriver.java
```diff
@@ -132,7 +132,8 @@ public void onNext(final ActiveContext activeContext) {
         LOG.log(Level.FINEST, "Submit group comm parameter server context {0} to {1}",
             new Object[]{nnCtrlCtxtId, contextId});
 
-        activeContext.submitContextAndService(groupCommContextConf, groupCommServiceConf);
+        activeContext.submitContextAndService(groupCommContextConf,
+            Configurations.merge(groupCommServiceConf, neuralNetworkESParameters.getServiceConfiguration()));
```
jsjason: What is this change for?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/data/NeuralNetworkDataParser.java
```diff
@@ -81,11 +79,10 @@ public void parse() {
         continue;
       }
       try {
-        final INDArray input = readNumpy(
-            new ByteArrayInputStream(text.getBytes()), delimiter);
-        final INDArray data = input.get(NDArrayIndex.interval(0, input.columns() - 2)).reshape(inputShape);
-        final int label = (int) input.getFloat(input.columns() - 2);
-        final boolean isValidation = ((int) input.getFloat(input.columns() - 1) == 1);
+        final Matrix input = readNumpy(matrixFactory, new ByteArrayInputStream(text.getBytes()), delimiter);
+        final Matrix data = input.get(range(0, input.getColumns() - 2));
```
jsjason: According to the comments on `range()`, the end index is exclusive. Shouldn't the second argument be `input.getColumns() - 1`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/data/NeuralNetworkDataParser.java
```diff
@@ -94,4 +91,26 @@ public void parse() {
...
+   * @param end the ending value, exclusive
+   * @return a generated array.
+   */
+  private int[] range(final int begin, final int end) {
+    if (begin > end) {
+      throw new IllegalArgumentException("The beginning value should be less than or equal to the ending value");
+    }
+    final int num = end - begin;
+    if (num == 0) {
+      return new int[]{begin};
```
jsjason: Shouldn't the return value be an empty array if `begin == end`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/FullyConnectedLayerParameterInitializer.java
```diff
@@ -57,15 +60,11 @@ public FullyConnectedLayerParameterInitializer(
   /** {@inheritDoc} */
   @Override
   public LayerParameter generateInitialParameter() {
-    final INDArray weight = Nd4j.randn(numInput, numOutput, randomSeed);
-    final INDArray bias = Nd4j.valueArrayOf(1, numOutput, initBias);
+    final Matrix weight = matrixFactory.randn(numInput, numOutput, randomSeed);
+    final Matrix bias = matrixFactory.zeros(numOutput).fill(initBias);
```
jsjason: We should use `matrixFactory.create()` instead of `zeros()` to avoid zero-initialization cost.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/provider/LocalNeuralNetParameterProvider.java
```diff
@@ -43,21 +43,28 @@
   private final LayerParameter[] deltaLayerParameters;
   private final float stepsize;
   private int numUpdate = 0;
+  private final MatrixFactory matrixFactory;
 
   @Inject
   public LocalNeuralNetParameterProvider(
-      @Parameter(SerializedLayerConfigurationSet.class) final Set<String> serializedLayerConfigurationSet,
+      @Parameter(SerializedLayerConfigurationSet.class)final Set<String> serializedLayerConfigurationSet,
```
jsjason: whitespace

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/provider/LocalNeuralNetParameterProvider.java
```diff
@@ -43,21 +43,28 @@
...
+    final Configuration matrixFactoryConf = Tang.Factory.getTang().newConfigurationBuilder()
+        .bindImplementation(MatrixFactory.class, matrixFactory.getClass())
+        .build();
 
     for (final String serializedInitializerConfiguration : serializedLayerConfigurationSet) {
       try {
         final Configuration initializerConfiguration =
             configurationSerializer.fromString(serializedInitializerConfiguration);
-        final Injector injector = Tang.Factory.getTang().newInjector(initializerConfiguration);
+        final Injector injector = Tang.Factory.getTang().newInjector(initializerConfiguration, matrixFactoryConf);
```
jsjason: Can we receive an Injector at the constructor and call `injector.forkInjector()` here? That way, we don't have to create `matrixFactoryConf`.

dolphin-dnn/src/test/java/edu/snu/dolphin/dnn/NeuralNetworkTest.java
```diff
@@ -15,45 +15,60 @@
...
+    }
+  }
 
-  private final INDArray input = Nd4j.create(new float[]{77, 57, 30, 26, 75, 74, 87, 75});
-  private final INDArray expectedOutput =
-      Nd4j.create(new float[]{5.96001904648e-01f, 5.54388443935e-01f, 4.88766051729e-01f});
-  private final INDArray label = Nd4j.create(new float[]{0, 1, 0});
+  private final Matrix input = matrixFactory.create(new float[]{77, 57, 30, 26, 75, 74, 87, 75});
+  private final Matrix expectedOutput =
+      matrixFactory.create(new float[]{5.61329743164e-01f, 5.97883503916e-01f, 4.79822915984e-01f});
```
jsjason: How did you compute these new test values?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/DefaultLayerParameterInitializer.java
```diff
@@ -29,18 +30,21 @@
 public final class DefaultLayerParameterInitializer implements LayerParameterInitializer {
 
   private final int index;
+  private final LayerParameter emptyLayerParam;
 
   @Inject
   public DefaultLayerParameterInitializer(
-      @Parameter(LayerConfigurationParameters.LayerIndex.class) final int index) {
+      final MatrixFactory matrixFactory,
+      @Parameter(LayerConfigurationParameters.LayerIndex.class)final int index) {
```
jsjason: whitespace

jsjason:
@beomyeol The new interface seems great. I left a few comments. Could you briefly explain why e5a6ca1 was needed?

beomyeol:
@jsjason I unintentionally added the configuration files that use a group communication and a parameter server. But, the commit that included this change did not state these configurations files. Since these files are useful, I made another separate commit for these and pushed it forcedly.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetworkGroupCommDriver.java
beomyeol: This change is for binding a matrix factory implementation which is used in the codecs for serializing and de-serializing `Matrix`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/data/NeuralNetworkDataParser.java
beomyeol: If the end index was `input.getColumns() - 1`, `data` would include the label value which is located in `input.getColumns() - 2`-th position.
So, the second argument should be `input.getColumns() - 2`.

dolphin-dnn/src/test/java/edu/snu/dolphin/dnn/NeuralNetworkTest.java
beomyeol: While `INDArray` uses a row-major array for storing elements, jblas's `FloatMatrix` uses a column-major array. Despite the same random generator and seed, this difference yields different randomly initialized matrices that represents fully connected layers' weights. This causes different outputs. Thus, I calculated the expected activations, errors and parameters with the new weight matrices using GNU Octave.

beomyeol:
e5a6ca1 is essential to compute the derivatives for the square function and the absolute function. The output values of these functions lose the signs of their input values. This change makes the computation of derivatives more expensive than before, because applying functions is needed one more time. However, this makes it possible to support all activation functions before the replacement of BLAS library.

beomyeol:
@jsjason I've address your comments and left some comments. Please take another look :)

jsjason:
I've tested this on several configurations and it works nicely. The code looks good, too. I'll merge it.

PR #149, opened 2015-12-10
Support batch processing for deep neural networks

### LINKED ISSUE #141: Support batch processing of DNN
In current implementation of `dolphin-dnn`, the neural network model is trained with only one input instance at each time. If we train the model with input batches, we can get better performance on computing activation values especially in fully connected layers since activations of multiple input instances are computed with one matrix multiplication.
Also, in model partitioning, we can split input batches, and each evaluator in the same model takes these splits and computes activation values and gradients.

beomyeol:
This pull request introduces batch processing for deep neural networks. A batch size that is used for training can be specified in the neural network configuration file. The changes involved in this PR are the followings.
* An input batch is given for training a neural network or inferring outputs.
* The signature of `push` for `ParameterProvider` is changed. The size of batch should be given.
* The return matrix for `generateParameterGradient` is the element-wise sum of parameter gradient matrices for each input instance in a batch.
* The messages that are used for reducing parameter gradients with group communication are changed from a list of layer parameters to a list of pairs of an integer (the size of batch) and a layer parameter (the element-wise sum of parameter gradients for an input batch).

This closes #141.

### DIFF
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetworkParameterUpdater.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetworkParameterUpdater.java
@@ -105,7 +105,7 @@ public final class NeuralNetworkParameterUpdater
    */
   private LayerParameter[] processLayerParameters(final String key,
                                                   final LayerParameter[] parameterGradients) {
-    if (parameterGradients == null || parameterGradients.length == 0 || !key.equals(WHOLE_MODEL)) {
+    if (parameterGradients == null || !key.equals(WHOLE_MODEL)) {
       return null;
     }
 
@@ -171,6 +171,10 @@ public final class NeuralNetworkParameterUpdater
    */
   private LayerParameter[] updateLayerParameter(final LayerParameter[] layerParameters,
                                                 final LayerParameter[] parameterGradients) {
+    if (layerParameters.length != parameterGradients.length) {
+      throw new RuntimeException("The number of parameter gradients is not equal to the number of layers.");
+    }
+
     for (int index = 0; index < layerParameters.length; ++index) {
       final LayerParameter layerParameter = layerParameters[index];
       final LayerParameter parameterGradient = parameterGradients[index];
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetworkParameterUpdater.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetworkParameterUpdater.java
@@ -15,7 +15,6 @@
  */
 package edu.snu.dolphin.dnn;
 
-import edu.snu.dolphin.dnn.blas.MatrixFactory;
 import edu.snu.dolphin.dnn.conf.NeuralNetworkConfigurationParameters;
 import edu.snu.dolphin.dnn.conf.NeuralNetworkConfigurationParameters.SerializedLayerConfigurationSet;
 import edu.snu.dolphin.dnn.data.NeuralNetParamServerData;
@@ -61,7 +60,6 @@ public final class NeuralNetworkParameterUpdater
   public static final String WHOLE_MODEL = "WHOLE_MODEL";
   public static final String VALIDATION = "VALIDATION";
 
-  private final MatrixFactory matrixFactory;
   private final Set<String> serializedLayerConfigurationSet;
   private final float stepsize;
   private final ConfigurationSerializer configurationSerializer;
@@ -71,13 +69,11 @@ public final class NeuralNetworkParameterUpdater
 
   @Inject
   private NeuralNetworkParameterUpdater(
-      final MatrixFactory matrixFactory,
       @Parameter(SerializedLayerConfigurationSet.class) final Set<String> serializedLayerConfigurationSet,
       @Parameter(NeuralNetworkConfigurationParameters.Stepsize.class) final float stepsize,
       final ConfigurationSerializer configurationSerializer,
       @Parameter(LogPeriod.class) final int logPeriod,
       final Injector injector) {
-    this.matrixFactory = matrixFactory;
     this.serializedLayerConfigurationSet = serializedLayerConfigurationSet;
     this.stepsize = stepsize;
     this.configurationSerializer = configurationSerializer;
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetworkParameterUpdater.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetworkParameterUpdater.java
@@ -15,7 +15,6 @@
  */
 package edu.snu.dolphin.dnn;
 
-import edu.snu.dolphin.dnn.blas.Matrix;
 import edu.snu.dolphin.dnn.blas.MatrixFactory;
 import edu.snu.dolphin.dnn.conf.NeuralNetworkConfigurationParameters;
 import edu.snu.dolphin.dnn.conf.NeuralNetworkConfigurationParameters.SerializedLayerConfigurationSet;
@@ -35,8 +34,6 @@ import org.apache.reef.tang.formats.ConfigurationSerializer;
 
 import javax.inject.Inject;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
 import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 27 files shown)


dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/data/IntAndLayerParameterArrayPairCodec.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  @Override
+  public byte[] encode(final Pair<Integer, LayerParameter[]> integerAndLayerParametersPair) {
+    try (final ByteArrayOutputStream bstream = new ByteArrayOutputStream();
+         final DataOutputStream dstream = new DataOutputStream(bstream)) {
+      encodeToStream(integerAndLayerParametersPair, dstream);
+      return bstream.toByteArray();
+    } catch (final IOException e) {
+      throw new RuntimeException("IOException during IntegerAndLayerParameterArrayPairCodec.encode()", e);
```
jsjason: The class name is different: `IntAndLayerParameterArrayPairCodec` / `IntegerAndLayerParameterArrayPairCodec`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/data/IntAndLayerParameterArrayPairListCodec.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  @Override
+  public byte[] encode(final List<Pair<Integer, LayerParameter[]>> intAndLayerParametersPairList) {
+    try (final ByteArrayOutputStream bstream = new ByteArrayOutputStream();
+         final DataOutputStream dstream = new DataOutputStream(bstream)) {
+      encodeToStream(intAndLayerParametersPairList, dstream);
+      return bstream.toByteArray();
+    } catch (final IOException e) {
+      throw new RuntimeException("IOException during IntegerAndLayerParameterArrayPairListCodec.encode()", e);
```
jsjason: The class name is different: `IntAndLayerParameterArrayPairListCodec ` / `IntegerAndLayerParameterArrayPairListCodec `?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/data/NeuralNetworkDataParser.java
```diff
@@ -83,13 +90,33 @@ public void parse() {
...
+        }
       } catch (final IOException e) {
         parseException = new ParseException("IOException: " + e.toString());
         return;
       }
     }
-    result = trainingData;
+
+    if (validationBatchGenerator.size() > 0) {
+      dataList.add(new Pair<>(validationBatchGenerator.pull(), false));
```
jsjason: Shouldn't the boolean value be `true`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/data/NeuralNetworkDataParser.java
```diff
@@ -83,13 +90,33 @@ public void parse() {
         final Matrix data = input.get(range(0, input.getColumns() - 2));
         final int label = (int) input.get(input.getColumns() - 2);
         final boolean isValidation = ((int) input.get(input.getColumns() - 1) == 1);
-        trainingData.add(new Pair<>(new Pair<>(data, label), isValidation));
+
+        if (isValidation) {
+          final Pair<Matrix, int[]> batch = validationBatchGenerator.push(data, label);
```
jsjason: Making `push` return a value seems a bit awkward. Does `BatchGenerator` have to be separated from this `NeuralNetworkDataParser` class, like the current code? Wouldn't be it okay to make `BatchGenerator` directly add pairs to `dataList` when the batch size is reached? The `matrixFactory` field of  `NeuralNetworkDataParser` could directly be used, too.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/data/NeuralNetworkDataParser.java
```diff
@@ -39,26 +41,29 @@
...
-  private List<Pair<Pair<Matrix, Integer>, Boolean>> result;
+  private final int batchSize;
+  private List<Pair<Pair<Matrix, int[]>, Boolean>> result;
   private ParseException parseException;
 
   @Inject
   private NeuralNetworkDataParser(final MatrixFactory matrixFactory,
                                   final DataSet<LongWritable, Text> dataSet,
-                                  @Parameter(Delimiter.class)final String delimiter) {
+                                  @Parameter(Delimiter.class)final String delimiter,
```
jsjason: We should've caught this before: whitespace between `.class)` and `final`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/provider/GroupCommParameterProvider.java
```diff
@@ -67,23 +56,17 @@ private GroupCommParameterProvider(
...
-      } catch (final NetworkException e) {
-        throw new RuntimeException("NetworkException while trying to send reduce", e);
-      } catch (final InterruptedException e) {
-        throw new RuntimeException("InterruptedException while trying to send reduce", e);
+  public synchronized void push(final int batchSize, final LayerParameter[] parameterGradients) {
+    try {
+      if (batchSize == 0 || parameterGradients == null || parameterGradients.length == 0) {
+        parameterGradientReduceSender.send(Collections.<Pair<Integer, LayerParameter[]>>emptyList());
+      } else {
+        parameterGradientReduceSender.send(Collections.singletonList(new Pair<>(batchSize, parameterGradients)));
```
jsjason: The group comm version now uses `List<Pair>`s as Reduce values, but the compute tasks send lists that contain only one object. Instead of sending `List<Pair>`s, how about sending the inner `Pair`s directly? This way, the summation of parameter gradients can be done across reducers instead of only at the controller task.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetworkParameterUpdater.java
```diff
@@ -103,47 +100,24 @@ public NeuralNetParamServerData process(final String key,
...
-   * Aggregate parameter gradients by computing the average of all gradients, per layer,
-   * and multiplying the step size.
+   * Multiplies the step size (in-place update).
    */
-  private List<LayerParameter[]> processLayerParametersList(final String key,
-                                                            final List<LayerParameter[]> parameterGradientsList) {
-    if (parameterGradientsList == null || parameterGradientsList.size() == 0 || !key.equals(WHOLE_MODEL)) {
+  private LayerParameter[] processLayerParameters(final String key,
+                                                  final LayerParameter[] parameterGradients) {
+    if (parameterGradients == null || parameterGradients.length == 0 || !key.equals(WHOLE_MODEL)) {
```
jsjason: Could we take out `parameterGradients.length == 0`? I know it's an unexpected condition, but its meaning is completely different from `parameterGradientsList.size() == 0`. Or, should we treat it just like any other error?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/data/NeuralNetworkDataParser.java
```diff
@@ -113,4 +140,75 @@ public void parse() {
...
+    private Matrix makeBatch(final List<Matrix> inputs) {
+      if (inputs.size() > 0) {
+        final Matrix ret = matrixFactory.create(inputs.size(), inputs.get(0).getLength());
+        int i = 0;
+        for (final Matrix vector : inputs) {
+          ret.putRow(i++, vector);
+        }
+        return ret;
+      } else {
+        throw new IllegalArgumentException("At least one ndarray is needed to make batch");
```
jsjason: Perhaps, "At least one **matrix**"?

jsjason:
@beomyeol Sorry for the late pass. I've left some comments for you. The sample dataset works well on both groupcomm and paramserver configurations.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/data/NeuralNetworkDataParser.java
beomyeol: As you mentioned, it will be some benefits not to have separate class for generating batches. However, without `BatchGenerator`, a code for data and label lists for training and validation inputs, and handling these lists, will look too complicated. That's why I introduced `BatchGenerator` to make code more clear.
Anyway, I will modify `push` method that returns a value and make `BatchGenerator` not static in order to use `matrixFactory` and `batchSize` directly.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/NeuralNetworkParameterUpdater.java
beomyeol: I'll remove a code for checking `parameterGradients.length == 0` at `processLayerParameters()` and add an error handling code for zero-length parameter gradients array at `updateLayerParameter()`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/provider/GroupCommParameterProvider.java
beomyeol: I tried to send `Pair` instead of `List<Pair>`. But, I encountered problems about how to define an empty message for synchronization and how a reducer handle this empty message for aggregation.

If we define `<0, null>` as an empty message, we should treat a `null` array at codecs.
If we define `<0, LayerParameter[0]>` as an empty message, there is an inconsistency for the number of layers and the number of layer parameters. (but, we may ignore this inconsistency.)
For `<0, LayerParameter[n]>` with `null` elements, we should treat `null` elements of an array at codecs.

With `List<Pair>`, the empty message is clear and easy to handle it at reducers. If you have a better idea, please let me know :)

beomyeol:
@jsjason Thank you for your review. I've addressed your comments and left a comment for the type of reduce values. Please take another look :)

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/provider/GroupCommParameterProvider.java
jsjason: In my opinion, the performance boost that can be gained by using `Pair`s and applying summation at reduce functions worths more than the code clearness we can achieve by using `List<Pair>`s. I think `<0, LayerParameter[0]>` is best for an empty message, because we can just ignore the length of the layer parameter array entirely if the batch size is zero.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/data/NeuralNetworkDataParser.java
```diff
@@ -83,13 +90,33 @@ public void parse() {
         final Matrix data = input.get(range(0, input.getColumns() - 2));
         final int label = (int) input.get(input.getColumns() - 2);
         final boolean isValidation = ((int) input.get(input.getColumns() - 1) == 1);
-        trainingData.add(new Pair<>(new Pair<>(data, label), isValidation));
+
+        if (isValidation) {
+          validationBatchGenerator.push(data, label);
+          if (validationBatchGenerator.check()) {
```
jsjason: I'm completely fine with having the inner class `BatchGenerator`, but I still think this code can be improved to be more concise. `check()` could be called inside `push`, and the next line `dataList.add(new Pair...)` could also be run inside `BatchGenerator` instead of here. I don't quite see why that would make the code more complicated.

jsjason:
@beomyeol I've done a second pass and left two comments. Thanks!

beomyeol:
@jsjason I've addressed your comments. Please review the changes.

jsjason:
Looks great, will merge.

PR #150, opened 2015-12-22
Remove required numInput and numOutput options for layer configuration

### LINKED ISSUE #146: Remove unnecessarily required num_input and num_output fields of LayerConfiguration in protocol buffer configurations
We require users to manually specifying the number of inputs and outputs for layers in neural network protocol buffer configuration files. However, the number of inputs can be determined by the number of outputs of the previous layer or `input_shape` (for the first layer), and the number of outputs can be determined by the number of inputs or layer-specific configurations (e.g. for activation layers, `num_output` is equal to `num_input` and for fully connected layers, `num_output` needs to be specified by the configuration). Thus, we can infer the number of inputs and outputs of layers without users' manual configurations.
Also, these manual configurations can cause inconsistency. For example, in the configuration for activation layers, users may set different values for `num_input` and `num_output`. To prevent this kind of situation, we should remove required `num_input` and `num_output` fields.

beomyeol:
This pull request removes `num_input` and `num_output` options for layer configurations. The `num_input` and `num_output` not only are meaningful only with 1-dimensional inputs, but also input and output shapes of layers can be inferred by other options for layers, such as kernel size for convolutional layers and the number of outputs for fully connected layers.

Moreover, the codes for instantiating layers and generating layer parameters are repeatedly used at several points. This PR introduces the utility class for neural network, `NeuralNetworkUtils`, for neural network related operations including generating layer instances and layer parameters.

In addition, this pull request brings back the support for arbitrary dimensional inputs that was lost in #148.
While input data had its dimension information with ND4J library, each layer has its input shape with which layers can process spatial operations on input data.

This closes #146.

### DIFF
--- a/dolphin-dnn/README.md
+++ b/dolphin-dnn/README.md
@@ -81,12 +81,12 @@ parameter_provider {
 
 **More types of layers such as convolutional layer and subsampling layer will be supported.**
 
-##### Activation Layers
+##### Activation Layer
 * Layer type: `Activation`
 * Parameters (`ActivationLayerConfiguration activation_param`)
 	* `activation_function`: the activation function to produce output values for this layer.
 
-##### Activation with Loss Layers
+##### Activation with Loss Layer
 * Layer type: `ActivationWithLoss`
 * Parameters (`ActivationWithLossLayerConfiguration activation_with_loss_param`)
 	* `activation_function`: the activation function to produce output values for this layer.
@@ -101,7 +101,7 @@ The following activation functions are supported.
 * Absolute: `abs`
 * Softmax: `softmax`
 
-##### Loss functions
+##### Loss Functions
 The following loss functions are supported.
 * CrossEntropy: `crossEntropy`
 
@@ -166,7 +166,7 @@ layer {
   type: "ActivationWithLoss"
   activation_with_loss_param {
     activation_function: "softmax"
-    loss_function: "crossentropy"
+    loss_function: "crossEntropy"
   }
 }
 
--- a/dolphin-dnn/README.md
+++ b/dolphin-dnn/README.md
@@ -164,8 +164,6 @@ layer {
 }
 layer {
   type: "ActivationWithLoss"
-  num_input: 10
-  num_output: 10
   activation_with_loss_param {
     activation_function: "softmax"
     loss_function: "crossentropy"
--- a/dolphin-dnn/README.md
+++ b/dolphin-dnn/README.md
@@ -76,19 +76,34 @@ parameter_provider {
 * Parameters (`FullyConnectedLayerConfiguration fully_connected_param`)
 	* `init_weight`: the standard deviation that is used to initialize the weights in this layer from a Gaussian distribution with mean 0.
 	* `init_bias`: constant value with which the biases of this layer are initialized.
-	* `activation_function`: the activation function to produce a output value for this layer.
+	* `random_seed`: the seed for generating random inital parameters.
+	* `num_output`: the number of outputs for this layer.
 
 **More types of layers such as convolutional layer and subsampling layer will be supported.**
 
-##### Activation function
-The following functions are supported.
+##### Activation Layers
+* Layer type: `Activation`
+* Parameters (`ActivationLayerConfiguration activation_param`)
+	* `activation_function`: the activation function to produce output values for this layer.
 
+##### Activation with Loss Layers
+* Layer type: `ActivationWithLoss`
+* Parameters (`ActivationWithLossLayerConfiguration activation_with_loss_param`)
+	* `activation_function`: the activation function to produce output values for this layer.
+	* `loss_function`: the loss function that is used to compute loss and calculate the loss gradient for backpropagation.
+
+##### Activation Functions
+The following activation functions are supported.
 * Sigmoid: `sigmoid`
 * ReLU: `relu`
 * TanhH: `tanh`
 * Power: `pow` (squared value)
 * Absolute: `abs`
 * Softmax: `softmax`
+
+##### Loss functions
+The following loss functions are supported.
+* CrossEntropy: `crossEntropy`
 
 ## How to run
 A script for training a neural network model is included with the source code, in `bin/run_neuralnetwork.sh`. `test/resources/data/neuralnet` is a sample subset of the [MNIST](http://yann.lecun.com/exdb/mnist) dataset, composed of 1,000 training images and 100 test images. `test/resources/configuration/neuralnet` is an example of a protocol buffer definition file; it defines a neural network model that uses two fully connected layers and a local parameter provider.
@@ -127,24 +142,36 @@ parameter_provider {
 }
 layer {
   type: "FullyConnected"
-  num_input: 784
-  num_output: 50
   fully_connected_param {
     init_weight: 1e-4
     init_bias: 2e-4
-    activation_function: "sigmoid"
+    num_output: 50
+  }
+}
+layer {
+  type: "Activation"
+  activation_param {
+    activation_function: "relu"
   }
 }
 layer {
   type: "FullyConnected"
-  num_input: 50
-  num_output: 10
   fully_connected_param {
     init_weight: 1e-2
     init_bias: 2e-2
-    activation_function: "sigmoid"
+    num_output: 10
   }
 }
+layer {
+  type: "ActivationWithLoss"
+  num_input: 10
+  num_output: 10
+  activation_with_loss_param {
+    activation_function: "softmax"
+    loss_function: "crossentropy"
+  }
+}
+
 ```
 This model comprises two fully connected layers with 50 and 10 features, respectively, and a local parameter provider. `input_shape` specifies the shape of input data. For the MNIST dataset, each data object is a 28 * 28 image and thus `input_shape` is configured as the following.
 
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/ActivationLayerConfigurationBuilder.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/ActivationLayerConfigurationBuilder.java
@@ -26,7 +26,7 @@ import org.apache.reef.util.Builder;
  * Configuration builder for activation layer.
  *
  * The configuration that this builder generates is used to create an activation layer instance.
- * The generate configuration need to bind the implementation for matrix factory, in order to inject layer instance.
+ * The generate configuration need to bind the parameter for a layer input shape, to inject layer instance.
  */
 public final class ActivationLayerConfigurationBuilder implements Builder<Configuration> {
 
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/ActivationLayerConfigurationBuilder.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/ActivationLayerConfigurationBuilder.java
@@ -26,6 +26,7 @@ import org.apache.reef.util.Builder;
  * Configuration builder for activation layer.
  *
  * The configuration that this builder generates is used to create an activation layer instance.
+ * The generate configuration need to bind the implementation for matrix factory, in order to inject layer instance.
  */
 public final class ActivationLayerConfigurationBuilder implements Builder<Configuration> {
 
@@ -33,19 +34,7 @@ public final class ActivationLayerConfigurationBuilder implements Builder<Config
     return new ActivationLayerConfigurationBuilder();
   }
 
-  private int numInput;
-  private int numOutput;
   private String activationFunction;
-
-  public synchronized ActivationLayerConfigurationBuilder setNumInput(final int numInput) {
-    this.numInput = numInput;
-    return this;
-  }
-
-  public synchronized ActivationLayerConfigurationBuilder setNumOutput(final int numOutput) {
-    this.numOutput = numOutput;
-    return this;
-  }
 
   public synchronized ActivationLayerConfigurationBuilder setActivationFunction(final String activationFunction) {
     this.activationFunction = activationFunction;
@@ -54,8 +43,6 @@ public final class ActivationLayerConfigurationBuilder implements Builder<Config
 
   public synchronized ActivationLayerConfigurationBuilder fromProtoConfiguration(
       final NeuralNetworkProtos.LayerConfiguration protoConf) {
-    numInput = protoConf.getNumInput();
-    numOutput = protoConf.getNumOutput();
     activationFunction = protoConf.getActivationParam().getActivationFunction();
     return this;
   }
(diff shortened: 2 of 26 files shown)


jsjason:
I'll do a pass today.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/ActivationLayerConfigurationBuilder.java
```diff
@@ -26,6 +26,7 @@
  * Configuration builder for activation layer.
  *
  * The configuration that this builder generates is used to create an activation layer instance.
+ * The generate configuration need to bind the implementation for matrix factory, in order to inject layer instance.
```
jsjason: Can you provide more details? I tried instantiating this class without any implementation bound to `MatrixFactory`, and I still was able to get an instance of `ActivationLayer`. Same for the other layers.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/LayerBase.java
```diff
@@ -39,13 +46,18 @@ public final int getIndex() {
   }
 
   /**
-   * @return the number of layer output nodes.
+   * @return the shape of inputs.
    */
-  public final int getNumOutput() {
-    return this.numOutput;
+  public final int[] getInputShape() {
```
jsjason: The only reason for having this public method and the one below seems to be to transform labels to one-hot vectors. Other than that, these methods look a bit redundant. Besides, the label-vector transformation can be done without the output shape known beforehand, by looking at the output themselves just before back propagation starts (around `NeuralNetwork.backPropagateTo`?). Am I missing something?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/util/NeuralNetworkUtils.java
```diff
@@ -0,0 +1,258 @@
...
+  private NeuralNetworkUtils() {
+  }
+
+  /**
+   * Delimiter that is used for distinguishing dimensions of shapes.
+   */
+  private static final String SHAPE_DELIMITER = ",";
+
+  /**
+   * Converts a list of integer for a shape to a string.
```
jsjason: typo: integers

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/LayerConfigurationParameters.java
```diff
@@ -43,14 +43,22 @@
...
-  @NamedParameter(doc = "number of layer input nodes", short_name = "numInput")
-  public static final class NumberOfInput implements Name<Integer> {
+  @NamedParameter(doc = "the shape of input data for layer")
+  public static final class LayerInputShape implements Name<String> {
   }
 
+  /**
+   * For fully connected layers.
+   */
+
```
jsjason: Redundant blank line. Same for the class below.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/util/NeuralNetworkUtils.java
```diff
@@ -0,0 +1,258 @@
...
+    }
+  }
+
+  /**
+   * De-serializes a set of serialized layer configurations to an array of layer configurations.
+   * @param configurationSerializer a configuration serializer to deserialize the specified configuration set.
+   * @param serializedLayerConfSet a set of serialized layer configurations.
+   * @return an array of layer configurations.
+   */
+  public static Configuration[] deserializeLayerConfSetToArray(
```
jsjason: This method and `getInitialLayerParameters` were originally one (not as an explicit method). What's the reason for separating them into two?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/util/NeuralNetworkUtils.java
```diff
@@ -0,0 +1,258 @@
...
+   * Converts a list of integer for a shape to a string.
+   * @param dimensionList a list of integers for a shape.
+   * @return a string for a shape.
+   */
+  public static String shapeToString(final List<Integer> dimensionList) {
+    return StringUtils.join(dimensionList, SHAPE_DELIMITER);
+  }
+
+  /**
+   * Converts an array of {@code int} for a shape to a string.
```
jsjason: typo: ints

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/NeuralNetworkConfigurationBuilder.java
```diff
@@ -35,25 +37,18 @@
  */
 public final class NeuralNetworkConfigurationBuilder implements Builder<Configuration> {
 
-  private Collection<String> layerConfigurations = new ArrayList<>();
-  private int index = 0;
+  private ArrayList<Configuration> layerConfigurations = new ArrayList<>();
```
jsjason: Any special reason to declare this variable as an `ArrayList`, and not just `List`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/LayerBase.java
```diff
@@ -17,18 +17,25 @@
...
  * Abstract class for the layer of a neural network.
  */
 public abstract class LayerBase {
 
   private final int index;
-  private final int numOutput;
   private LayerParameter layerParameter;
+  private final int[] inputShape;
+
+  protected LayerBase(final int index, final int[] inputShape) {
```
jsjason: This constructor is not being used anywhere. Do we need this?

jsjason:
I finished my pass, @beomyeol.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/ActivationLayerConfigurationBuilder.java
beomyeol: Sorry for confusing you. They need to bind `LayerInputShape`, not `matrixFactory`. I'll fix it.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/LayerBase.java
beomyeol: Generating one-hot output vectors was the main reason why `getNumOutput()` exists. However, `getOutputShape()` is needed to provide information about the shape of inputs to subsequent layers at `getLayerInstances()` in `NeuralNetworkUtils`. Actually, `getInputShape()` does not have to be public. But, it may offer more information about layers.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/util/NeuralNetworkUtils.java
beomyeol: De-serializing a set of serialized layer configuration is the common part for `getInitialLayerParameters()` and `getLayerInstance()`. Making this part as a method can remove the repeated codes.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/NeuralNetworkConfigurationBuilder.java
beomyeol: At `build()`, we get the i-th configuration, using `layerConfiguartions.get()`. If we declare this variable as a `List`, the performance `get()` function may be O(n). However, declaring it as an `ArrayList` guarantees O(1) performance. Although it may be the very little performance gain,
This seems to be fine since it is a private member variable,

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/LayerBase.java
beomyeol: This constructor is used in constructors in all layers. `super(index, inputShape)`.

bgchun:
@beomyeol @jsjason Is this PR ready to be merged?

jsjason:
@bgchun @beomyeol Sorry, I didn't have the chance to do another pass up till now. I'll do a review right away.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/NeuralNetworkConfigurationBuilder.java
jsjason: This variable is guaranteed to have O(1) `get()` performance even if you declare it as a `List`, because you're initializing it with `new ArrayList<>()`. We should follow the convention of using interfaces instead of concrete classes for declaring variables, when we can.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/util/NeuralNetworkUtils.java
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  /**
+   * Returns initial layer parameters from the specified set of configurations and the input shape.
+   * @param layerInitializerConfs an array of configurations for injecting layer initializers
+   * @param inputShape an input shape for the neural network.
+   * @param matrixFactoryClass a matrix factory class for instantiate matrices
+   * @return an array of initial layer parameters
+   */
+  public static LayerParameter[] getInitialLayerParameters(final Configuration[] layerInitializerConfs,
```
jsjason: This method isn't being used.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/util/NeuralNetworkUtils.java
```diff
@@ -0,0 +1,258 @@
...
+  }
+
+  /**
+   * Returns layer instances from the specified array of configurations for layer instances.
+   * @param layerConfs an array of configurations for layer instances
+   * @param inputShape an input shape for the neural network
+   * @param matrixFactoryClass a class for {@link MatrixFactory}
+   * @return an array of layer instances
+   */
+  public static LayerBase[] getLayerInstances(final Configuration[] layerConfs,
```
jsjason: This method isn't being used.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/LayerBase.java
jsjason: I did a 'Find Usages' check on this constructor and IntelliJ says no usages were found. You're probably talking about the other constructor, `protected LayerBase(final int index, final String inputShapeString)`. I think this unused one can be removed.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/util/NeuralNetworkUtils.java
```diff
@@ -0,0 +1,258 @@
...
+   * @param inputShape an input shape for the neural network.
+   * @return an array of initial layer parameters
+   */
+  public static LayerParameter[] getInitialLayerParameters(final Injector injector,
+                                                           final Configuration[] layerInitializerConfs,
+                                                           final String inputShape) {
+    final LayerParameter[] layerParameters = new LayerParameter[layerInitializerConfs.length];
+
+    String currentInputShape = inputShape;
+    for (int i = 0; i < layerInitializerConfs.length; ++i) {
```
jsjason: We can use foreach here.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/util/NeuralNetworkUtils.java
jsjason: I get your point, but as a result we're iterating over the layer configurations twice instead of once per one `getInitialLayerParameters()` or `getLayerInstances()` call. I guess it wouldn't be a big problem since the number of layers isn't so big, though..

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/util/NeuralNetworkUtils.java
```diff
@@ -0,0 +1,258 @@
...
+   * @param inputShape an input shape for the neural network.
+   * @return an array of layer instances.
+   */
+  public static LayerBase[] getLayerInstances(final Injector injector,
+                                              final Configuration[] layerConfs,
+                                              final String inputShape) {
+    final LayerBase[] layers = new LayerBase[layerConfs.length];
+
+    String currentInputShape = inputShape;
+    for (int i = 0; i < layerConfs.length; ++i) {
```
jsjason: We can use foreach here.

jsjason:
@beomyeol I did another pass. I'll merge this right away after you check my comments. Thanks.

beomyeol:
@jsjason Thank you for your review. I've addressed your comments and pushed one more commit for updating README. Please take another look :)

dolphin-dnn/README.md
```diff
@@ -76,13 +76,24 @@ parameter_provider {
...
 	* `init_bias`: constant value with which the biases of this layer are initialized.
-	* `activation_function`: the activation function to produce a output value for this layer.
+	* `random_seed`: the seed for generating random inital parameters.
+	* `num_output`: the number of outputs for this layer.
 
 **More types of layers such as convolutional layer and subsampling layer will be supported.**
 
-##### Activation function
-The following functions are supported.
+##### Activation Layers
```
jsjason: `Activation Layer`, without the trailing s.

dolphin-dnn/README.md
```diff
@@ -90,6 +101,10 @@ The following functions are supported.
 * Absolute: `abs`
 * Softmax: `softmax`
 
+##### Loss functions
+The following loss functions are supported.
+* CrossEntropy: `crossEntropy`
```
jsjason: This should be `crossentropy`.

dolphin-dnn/README.md
```diff
@@ -76,13 +76,24 @@ parameter_provider {
...
 **More types of layers such as convolutional layer and subsampling layer will be supported.**
 
-##### Activation function
-The following functions are supported.
+##### Activation Layers
+* Layer type: `Activation`
+* Parameters (`ActivationLayerConfiguration activation_param`)
+	* `activation_function`: the activation function to produce output values for this layer.
 
+##### Activation with Loss Layers
```
jsjason: `Activation with Loss Layer`, without the trailing s.

dolphin-dnn/README.md
```diff
@@ -90,6 +101,10 @@ The following functions are supported.
 * Absolute: `abs`
 * Softmax: `softmax`
 
+##### Loss functions
```
jsjason: Let's use consistent capitalization: `Loss Functions`.

jsjason:
@beomyeol I did another pass for the new README commit you added. Thanks.

dolphin-dnn/README.md
beomyeol: Letter case in configuration file does not matter since our code is case-insensitive for strings in the configuration file. `crossEntropy` is better to read than `crossentropy`. I'll change configuration samples to use `crossEntropy`.

beomyeol:
@jsjason Thank you for your comments. I've address them.

jsjason:
@beomyeol Great. I'll merge.

PR #163, opened 2016-01-18
Use batch matrices in which each column represents an input instance.

### LINKED ISSUE #155: Save input instances in columns of input batch matrices
For batch processing, `dolphin-dnn` deals with matrices in which each row represents one input instance. However, `Matrix` interface and jblas library on which `MatrixJBLASImpl` depends, use column-major arrays for saving elements. Because of this inconsistency, some operations that can be commonly used, will be inefficient.
For example, getting one instance (row) in a batch matrix should access and concatenate not contiguous parts of the inner array. 
Matlab, Fortran libraries, and other BLAS libraries including JBLAS and Breeze use column-major array. Thus, we should consider saving each instance in columns of input batch matrices.

beomyeol:
For batch processing, `dolphin-dnn` handles matrices in which each **row** represents an input instance. However, many linear algebra libraries such as [netlib-java](https://github.com/fommil/netlib-java), [jblas](http://jblas.org), and [cuBLAS](https://developer.nvidia.com/cublas) use column-major storage for the compatibility with Fortran environment.
In this circumstance, retrieving each input instance from batch matrices requires the access of not contiguous memory (sequential access), which is inefficient.
This pull request changes the batch matrix format that each **column** represents an input instance, not each **row**. This PR also include the change that `MatrixFactory` creates column vectors when it generates vectors.

This closes #155

### DIFF
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixUtils.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixUtils.java
@@ -123,7 +123,7 @@ public final class MatrixUtils {
   }
 
   /**
-   * Creates a column vector in which only element at the specified position is {@code 1}
+   * Creates a column vector in which the only element at the specified position is {@code 1}
    * and other elements are {@code 0}.
    *
    * @param matrixFactory a matrix factory used to create a matrix
@@ -136,7 +136,7 @@ public final class MatrixUtils {
   }
 
   /**
-   * Creates a matrix of which each column is an one-hot vector specified by each element of the given indices array.
+   * Creates a matrix of which each column is a one-hot vector specified by each element of the given indices array.
    * @param matrixFactory a matrix factory used to create a matrix
    * @param indices the array of indices that indicate the one-hot positions
    * @param length the length of each column vector, in other words, the number of rows of the return matrix
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixUtils.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixUtils.java
@@ -79,23 +79,23 @@ public final class MatrixUtils {
     final BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream));
     String line;
     final List<float[]> dataList = new ArrayList<>();
-    int numColumns = -1;
+    int numRows = -1;
     final Matrix ret;
     while ((line = reader.readLine()) != null) {
       final String[] data = line.trim().split(delimiter);
-      if (numColumns < 0) {
-        numColumns = data.length;
+      if (numRows < 0) {
+        numRows = data.length;
       } else {
-        if (data.length != numColumns) {
-          throw new RuntimeException("Data has inconsistent number of columns");
+        if (data.length != numRows) {
+          throw new RuntimeException("Data has inconsistent length");
         }
       }
       dataList.add(readSplit(data));
     }
 
-    ret = matrixFactory.create(dataList.size(), numColumns);
+    ret = matrixFactory.create(numRows, dataList.size());
     for (int i = 0; i < dataList.size(); i++) {
-      ret.putRow(i, matrixFactory.create(dataList.get(i)));
+      ret.putColumn(i, matrixFactory.create(dataList.get(i)));
     }
     return ret;
   }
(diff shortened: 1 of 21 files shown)


jsjason:
I'll do pass today.

seyounglee95:
NAVER - http://www.naver.com/
--------------------------------------------

fatalu1226@naver.com 님께 보내신 메일 <Re: [dolphin] Use batch matrices in which each column represents an input instance. (#163)> 이 다음과 같은 이유로 전송 실패했습니다.

--------------------------------------------

받는 사람이 회원님의 메일을 수신차단 하였습니다. 


--------------------------------------------

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixUtils.java
```diff
@@ -123,28 +123,29 @@ public static Matrix readNumpy(final MatrixFactory matrixFactory,
   }
 
   /**
-   * Creates a row vector in which only element at the specified position is {@code 1} and other elements are {@code 0}.
+   * Creates a column vector in which only element at the specified position is {@code 1}
```
jsjason: "only **the** element at"

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/MatrixUtils.java
```diff
@@ -123,28 +123,29 @@ public static Matrix readNumpy(final MatrixFactory matrixFactory,
...
+   * @param length the length of a column vector
+   * @return a generated column vector
    */
   public static Matrix createOutputVector(final MatrixFactory matrixFactory, final int index, final int length) {
     return matrixFactory.zeros(length).put(index, 1.0f);
   }
 
   /**
-   * Creates a matrix of which each row is a one-hot vector specified by each element of the given indices array.
+   * Creates a matrix of which each column is an one-hot vector specified by each element of the given indices array.
```
jsjason: "is a~~n~~ one-hot vector"

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/blas/function/Softmax.java
```diff
@@ -20,6 +20,8 @@
 
 /**
  * Softmax function.
+ *
+ * Assumes each column represents each input instance.
```
jsjason: "represents ~~each~~**a single** input instance"

dolphin-dnn/src/test/java/edu/snu/dolphin/dnn/blas/function/AbsoluteTest.java
```diff
@@ -44,10 +44,19 @@ public void setup() throws InjectionException {
...
-    this.input = matrixFactory.create(new float[][]{{1.0f, -2.0f, 3.0f}, {-4.0f, 5.0f, -6.0f}});
-    this.expectedOutput = matrixFactory.create(new float[][]{{1.0f, 2.0f, 3.0f}, {4.0f, 5.0f, 6.0f}});
-    this.expectedDerivative = matrixFactory.create(new float[][]{
-        {1.0f, -1.0f, 1.0f}, {-1.0f, 1.0f, -1.0f}});
+    final int numInput = 3;
+    final int numBatch = 2;
+    this.input = matrixFactory.create(new float[]{
+        1.0f, -2.0f, 3.0f,
+        -4.0f, 5.0f, -6.0f},
+        numInput, numBatch);
```
jsjason: What's the reason for using `create(float[], int, int)` and not `create(float[][])` like before? There seems to similar changes in other test classes.

dolphin-dnn/src/test/java/edu/snu/dolphin/dnn/NeuralNetworkTest.java
```diff
@@ -58,17 +58,17 @@
 
   private final Matrix input = matrixFactory.create(new float[]{77, 57, 30, 26, 75, 74, 87, 75});
   private final Matrix expectedOutput =
-      matrixFactory.create(new float[]{5.61329743164e-01f, 5.97883503916e-01f, 4.79822915984e-01f});
+      matrixFactory.create(new float[]{5.96001906923e-01f, 5.54388444438e-01f, 4.88766051614e-01f});
```
jsjason: I thought that only the orientation of matrices would be changed in the tests (row/column), but it seems that many values in the tests have been changed too. What's the reason?

jsjason:
I've left a few comments, @beomyeol. Thanks.

dolphin-dnn/src/test/java/edu/snu/dolphin/dnn/NeuralNetworkTest.java
beomyeol: The shapes of fully connected layers' weight matrices are changed from `numInput x numOutput` to `numOutput x numInput`. Thus, the positions of generated random numbers in the matrix are changed, which generates different outputs.
FYI, if using weight matrices generated by `matrixFactory.randn(numInput, numOutput, randomSeed).transpose()` in `FullyConnectedLayerParameterInitializer`, we can get the same results. But, we should avoid using it because `transpose()` is a costly function.

dolphin-dnn/src/test/java/edu/snu/dolphin/dnn/blas/function/AbsoluteTest.java
beomyeol: Since a single input instance is stored in each column, it seems not straightforward in Java representation if we use `create(float[][])`.
For example, we should use `create(new float[][]{{1.0f, -4.0f}, {-2.0f, 5.0f}, {3.0f, -6.0f}}` for generating `input` that consists of two input instances `(1, -2, 3)` and `(-4, 5, -6)`.
However, multiple instances can be distinguishably represented with line breaks when we use `create(float[], int, int)`.

beomyeol:
@jsjason Thank you for the review. I've addressed some comments. Also, I left comments, too.

jsjason:
Looks great. The sample configuration/dataset works well too. I'll merge this.

PR #164, opened 2016-01-19
pooling layer base implementation

seyounglee95:
As a first step for implementing CNN,  this PR introduce new layer class `PoolingLayer`.  This layer extends `LayerBase`, and include implementation for `isLearnable()`, `getOutputShape()` and `generateParameterGradient()`. This PR include some modifications in `LayerParameter`, which enable `poolingLayer`  to make use of its new parameters. Also, `PoolingLayerConfigurationBuilder` class and `PoolingLayerParameterInitializer` class are added  for configuration of pooling layer. `neural_network.proto` is modified to take inputs for pooling layer. Concrete implementation for `feedForward()` and `backPropagate()` will be made in later PRs.

### DIFF
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/LayerConfigurationParameters.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/LayerConfigurationParameters.java
@@ -64,4 +64,23 @@ public final class LayerConfigurationParameters {
   @NamedParameter(doc = "loss function of loss layer", short_name = "lossFunc")
   public static final class LossFunction implements Name<String> {
   }
+
+  /**
+   * For pooling layers.
+   */
+  @NamedParameter(doc = "pooling type of pooling layer", short_name = "poolingTyp")
+  public static final class PoolingType implements Name<String> {
+  }
+
+  @NamedParameter(doc = "stride of pooling / convolutional layer", short_name = "stride")
+  public static final class Stride implements Name<Integer> {
+  }
+
+  @NamedParameter(doc = "kernel height of pooling / convolutional layer", short_name = "height")
+  public static final class KernelHeight implements Name<Integer> {
+  }
+
+  @NamedParameter(doc = "kernal width of pooling / convolutional layer", short_name = "width")
+  public static final class KernelWidth implements Name<Integer> {
+  }
 }
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/LayerConfigurationParameters.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/LayerConfigurationParameters.java
@@ -72,15 +72,19 @@ public final class LayerConfigurationParameters {
   public static final class PoolingType implements Name<String> {
   }
 
-  @NamedParameter(doc = "stride of pooling / convolutional layer", short_name = "stride")
-  public static final class Stride implements Name<Integer> {
+  @NamedParameter(doc = "stride height of pooling / convolutional layer", short_name = "strideH")
+  public static final class StrideHeight implements Name<Integer> {
   }
 
-  @NamedParameter(doc = "kernel height of pooling / convolutional layer", short_name = "height")
+  @NamedParameter(doc = "stride width of pooling / convolutional layer", short_name = "strideW")
+  public static final class StrideWidth implements Name<Integer> {
+  }
+
+  @NamedParameter(doc = "kernel height of pooling / convolutional layer", short_name = "kernelH")
   public static final class KernelHeight implements Name<Integer> {
   }
 
-  @NamedParameter(doc = "kernal width of pooling / convolutional layer", short_name = "width")
+  @NamedParameter(doc = "kernel width of pooling / convolutional layer", short_name = "kernelW")
   public static final class KernelWidth implements Name<Integer> {
   }
 }
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/LayerConfigurationParameters.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/LayerConfigurationParameters.java
@@ -68,7 +68,7 @@ public final class LayerConfigurationParameters {
   /**
    * For pooling layers.
    */
-  @NamedParameter(doc = "pooling type of pooling layer", short_name = "poolingTyp")
+  @NamedParameter(doc = "pooling type of pooling layer")
   public static final class PoolingType implements Name<String> {
   }
 
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/PoolingLayerConfigurationBuilder.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/PoolingLayerConfigurationBuilder.java
@@ -28,7 +28,7 @@ import org.apache.reef.util.Builder;
  * Configuration builder for Pooling layer.
  *
  * The configuration that this builder generates is used to create a pooling layer instance.
- * The generate configuration need to bind the parameter for a layer input shape, to inject layer instance.
+ * The generated configuration needs to bind the parameter for a layer input shape, to inject a layer instance.
  */
 public final class PoolingLayerConfigurationBuilder implements Builder<Configuration> {
 
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/PoolingLayerConfigurationBuilder.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/PoolingLayerConfigurationBuilder.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright (C) 2016 Seoul National University
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package edu.snu.dolphin.dnn.conf;
+
+import edu.snu.dolphin.dnn.layerparam.initializer.LayerParameterInitializer;
+import edu.snu.dolphin.dnn.layerparam.initializer.PoolingLayerParameterInitializer;
+import edu.snu.dolphin.dnn.layers.PoolingLayer;
+import edu.snu.dolphin.dnn.layers.LayerBase;
+import edu.snu.dolphin.dnn.proto.NeuralNetworkProtos;
+import org.apache.reef.tang.Configuration;
+import org.apache.reef.tang.Tang;
+import org.apache.reef.util.Builder;
+
+/**
+ * Configuration builder for Pooling layer.
+ *
+ * The configuration that this builder generates is used to create a pooling layer instance.
+ * The generate configuration need to bind to bind the parameter for a layer input shape, to inject layer instance.
+ */
+public final class PoolingLayerConfigurationBuilder implements Builder<Configuration> {
+
+  public static PoolingLayerConfigurationBuilder newConfigurationBuilder() {
+    return new PoolingLayerConfigurationBuilder();
+  }
+
+  private String poolingType;
+  private int strideHeight;
+  private int strideWidth;
+  private int kernelHeight;
+  private int kernelWidth;
+
+  public synchronized PoolingLayerConfigurationBuilder setPoolingType(final String poolingType) {
+    this.poolingType = poolingType;
+    return this;
+  }
+
+  public synchronized PoolingLayerConfigurationBuilder setStrideHeight(final int strideHeight) {
+    this.strideHeight = strideHeight;
+    return this;
+  }
+
+  public synchronized PoolingLayerConfigurationBuilder setStrideWidth(final int strideWidth) {
+    this.strideWidth = strideWidth;
+    return this;
+  }
+
+  public synchronized PoolingLayerConfigurationBuilder setKernelHeight(final int kernelHeight) {
+    this.kernelHeight = kernelHeight;
+    return this;
+  }
+
+  public synchronized PoolingLayerConfigurationBuilder setKernelWidth(final int kernelWidth) {
+    this.kernelWidth = kernelWidth;
+    return this;
+  }
+
+  public synchronized PoolingLayerConfigurationBuilder fromProtoConfiguration(
+      final NeuralNetworkProtos.LayerConfiguration protoConf) {
+    poolingType = protoConf.getPoolingParam().getPoolingType();
+    strideHeight = protoConf.getPoolingParam().getStrideHeight();
+    strideWidth = protoConf.getPoolingParam().getStrideWidth();
+    kernelHeight = protoConf.getPoolingParam().getKernelHeight();
+    kernelWidth = protoConf.getPoolingParam().getKernelWidth();
+    return this;
+  }
+
+  @Override
+  public synchronized Configuration build() {
+    return Tang.Factory.getTang().newConfigurationBuilder()
+        .bindNamedParameter(LayerConfigurationParameters.PoolingType.class, String.valueOf(poolingType))
+        .bindNamedParameter(LayerConfigurationParameters.StrideHeight.class, String.valueOf(strideHeight))
+        .bindNamedParameter(LayerConfigurationParameters.StrideWidth.class, String.valueOf(strideWidth))
+        .bindNamedParameter(LayerConfigurationParameters.KernelHeight.class, String.valueOf(kernelHeight))
+        .bindNamedParameter(LayerConfigurationParameters.KernelWidth.class, String.valueOf(kernelWidth))
+        .bindImplementation(LayerBase.class, PoolingLayer.class)
+        .bindImplementation(LayerParameterInitializer.class, PoolingLayerParameterInitializer.class)
+        .build();
+  }
+}
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/PoolingLayerConfigurationBuilder.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/PoolingLayerConfigurationBuilder.java
@@ -28,7 +28,7 @@ import org.apache.reef.util.Builder;
  * Configuration builder for Pooling layer.
  *
  * The configuration that this builder generates is used to create a pooling layer instance.
- * The generate configuration need to bind to bind the parameter for a layer input shape, to inject layer instance.
+ * The generate configuration need to bind the parameter for a layer input shape, to inject layer instance.
  */
 public final class PoolingLayerConfigurationBuilder implements Builder<Configuration> {
 
@@ -36,9 +36,9 @@ public final class PoolingLayerConfigurationBuilder implements Builder<Configura
     return new PoolingLayerConfigurationBuilder();
   }
 
-  private String poolingType;
-  private int strideHeight;
-  private int strideWidth;
+  private String poolingType = "MAX";
+  private int strideHeight = 1;
+  private int strideWidth = 1;
   private int kernelHeight;
   private int kernelWidth;
 
@@ -69,9 +69,15 @@ public final class PoolingLayerConfigurationBuilder implements Builder<Configura
 
   public synchronized PoolingLayerConfigurationBuilder fromProtoConfiguration(
       final NeuralNetworkProtos.LayerConfiguration protoConf) {
-    poolingType = protoConf.getPoolingParam().getPoolingType();
-    strideHeight = protoConf.getPoolingParam().getStrideHeight();
-    strideWidth = protoConf.getPoolingParam().getStrideWidth();
+    if (protoConf.getPoolingParam().hasPoolingType()) {
+      poolingType = protoConf.getPoolingParam().getPoolingType();
+    }
+    if (protoConf.getPoolingParam().hasStrideHeight()) {
+      strideHeight = protoConf.getPoolingParam().getStrideHeight();
+    }
+    if (protoConf.getPoolingParam().hasStrideWidth()) {
+      strideWidth = protoConf.getPoolingParam().getStrideWidth();
+    }
     kernelHeight = protoConf.getPoolingParam().getKernelHeight();
     kernelWidth = protoConf.getPoolingParam().getKernelWidth();
     return this;
@@ -80,7 +86,7 @@ public final class PoolingLayerConfigurationBuilder implements Builder<Configura
   @Override
   public synchronized Configuration build() {
     return Tang.Factory.getTang().newConfigurationBuilder()
-        .bindNamedParameter(LayerConfigurationParameters.PoolingType.class, String.valueOf(poolingType))
+        .bindNamedParameter(LayerConfigurationParameters.PoolingType.class, poolingType)
         .bindNamedParameter(LayerConfigurationParameters.StrideHeight.class, String.valueOf(strideHeight))
         .bindNamedParameter(LayerConfigurationParameters.StrideWidth.class, String.valueOf(strideWidth))
         .bindNamedParameter(LayerConfigurationParameters.KernelHeight.class, String.valueOf(kernelHeight))
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/PoolingLayerConfigurationBuilder.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/PoolingLayerConfigurationBuilder.java
@@ -36,9 +36,9 @@ public final class PoolingLayerConfigurationBuilder implements Builder<Configura
     return new PoolingLayerConfigurationBuilder();
   }
 
-  private String poolingType = "MAX";
-  private int strideHeight = 1;
-  private int strideWidth = 1;
+  private String poolingType;
+  private int strideHeight;
+  private int strideWidth;
   private int kernelHeight;
   private int kernelWidth;
 
@@ -69,15 +69,9 @@ public final class PoolingLayerConfigurationBuilder implements Builder<Configura
 
   public synchronized PoolingLayerConfigurationBuilder fromProtoConfiguration(
       final NeuralNetworkProtos.LayerConfiguration protoConf) {
-    if (protoConf.getPoolingParam().hasPoolingType()) {
-      poolingType = protoConf.getPoolingParam().getPoolingType();
-    }
-    if (protoConf.getPoolingParam().hasStrideHeight()) {
-      strideHeight = protoConf.getPoolingParam().getStrideHeight();
-    }
-    if (protoConf.getPoolingParam().hasStrideWidth()) {
-      strideWidth = protoConf.getPoolingParam().getStrideWidth();
-    }
+    poolingType = protoConf.getPoolingParam().getPoolingType();
+    strideHeight = protoConf.getPoolingParam().getStrideHeight();
+    strideWidth = protoConf.getPoolingParam().getStrideWidth();
     kernelHeight = protoConf.getPoolingParam().getKernelHeight();
     kernelWidth = protoConf.getPoolingParam().getKernelWidth();
     return this;
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/PoolingLayerParameterInitializer.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/PoolingLayerParameterInitializer.java
@@ -27,12 +27,13 @@ import static edu.snu.dolphin.dnn.util.NeuralNetworkUtils.shapeFromString;
 /**
  * Pooling Layer parameter initializer.
  *
- * This initializer is for pooling layer which do not have layer parameter.
+ * This initializer is for pooling layers which do not have layer parameters.
  */
 public final class PoolingLayerParameterInitializer implements LayerParameterInitializer {
 
   private final int index;
   private final int[] inputShape;
+  private final int[] outputShape;
   private final int strideHeight;
   private final int strideWidth;
   private final int kernelHeight;
@@ -40,7 +41,7 @@ public final class PoolingLayerParameterInitializer implements LayerParameterIni
   private final LayerParameter emptyLayerParam;
 
   @Inject
-  public PoolingLayerParameterInitializer(
+  private PoolingLayerParameterInitializer(
       final MatrixFactory matrixFactory,
       @Parameter(LayerConfigurationParameters.LayerIndex.class) final int index,
       @Parameter(LayerConfigurationParameters.LayerInputShape.class) final String inputShape,
@@ -54,6 +55,7 @@ public final class PoolingLayerParameterInitializer implements LayerParameterIni
     this.strideWidth = strideWidth;
     this.kernelHeight = kernelHeight;
     this.kernelWidth = kernelWidth;
+    this.outputShape = computeOutputShape();
     this.emptyLayerParam = LayerParameter.newEmptyInstance(matrixFactory);
   }
 
@@ -75,24 +77,32 @@ public final class PoolingLayerParameterInitializer implements LayerParameterIni
    * This function computes output shape.
    * input shape: row * col
    * output shape: row' * col'
-   * row = (row − kernelHeight) / stride + 1
-   * col = (col − kernelWidth) / stride + 1
+   * row' = (row − kernelHeight) / stride + 1
+   * col' = (col − kernelWidth) / stride + 1
+   * @return shape of output
    */
-  @Override
-  public int[] getOutputShape() {
+  private int[] computeOutputShape() {
     final int[] computedShape;
     switch (inputShape.length) {
-    case 1 :
+    case 1:
       computedShape = new int[1];
       computedShape[0] = (inputShape[0] - kernelHeight) / strideHeight + 1;
       return computedShape;
-    case 2 :
+    case 2:
       computedShape = new int[2];
       computedShape[0] = (inputShape[0] - kernelHeight) / strideHeight + 1;
       computedShape[1] = (inputShape[1] - kernelWidth) / strideWidth + 1;
       return computedShape;
-    default :
+    default:
       throw new IllegalArgumentException("Unsupported input dimension: " + Integer.toString(inputShape.length));
     }
+  }
+
+  /**
+   * @return shape of output
+   */
+  @Override
+  public int[] getOutputShape() {
+    return outputShape;
   }
 }
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/PoolingLayerParameterInitializer.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/PoolingLayerParameterInitializer.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright (C) 2016 Seoul National University
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package edu.snu.dolphin.dnn.layerparam.initializer;
+
+import edu.snu.dolphin.dnn.blas.MatrixFactory;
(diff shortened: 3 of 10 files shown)


beomyeol:
@seyounglee95 I'll take a look. However, test build failed since `NullPointException` is thrown during `LayerParameterArrayCodecTest`. Please fix this :)

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/LayerParameter.java
```diff
@@ -24,6 +24,10 @@
 public final class LayerParameter {
   private final Matrix weightParam;
   private final Matrix biasParam;
+  private final String poolingType;
+  private final int stride;
+  private final int kernalHeight;
+  private final int kernalWidth;
```
beomyeol: These parameters should not be in `LayerParameter`. `LayerParameter` is used for storing weight and bias. These parameters can be member variables of `PoolingLayer`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,144 @@
+/*
+ * Copyright (C) 2015 Seoul National University
```
beomyeol: Please change `2015` to `2016`

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,144 @@
...
+
+import javax.inject.Inject;
+
+/**
+ * Pooling layer.
+ *
+ * This layer is not learnable.
+ * This layer resizes input matrix spatially, using max pooling or average pooling.
+ * In a forward pass,
+ * max pooling picks maximum value in certain range ( kernalHeight * kernalWidth) and these values make up output.
```
beomyeol: Please remove a space right after `(` in `( kernalHeight * kernalWidth)`.

beomyeol: Typo `kernel`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,144 @@
...
+ * average pooling get average of values in certain range (kernalHeight * kernalWidth) and these values make up output.
```
beomyeol: Please capitalize the first letter and fix `get` to `gets`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,144 @@
...
+ * In a backward pass,
+ * each value of error matrix is the sum of elements in next error matrix
+ * on which had an impact in feedforward step.
+ */
+
+public abstract class PoolingLayer extends LayerBase {
```
beomyeol: Please remove `abstract` keyword and add `final`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,144 @@
...
+
+  private final int[] outputShape;
+  private Matrix trackMatrix;
+  private Function poolingFunctions;
+
+  @Inject
+  public PoolingLayer(@Parameter(LayerConfigurationParameters.LayerIndex.class) final int index,
```
beomyeol: Please make this constructor `private`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,144 @@
...
+  /**
+   * Computes output values for this pooling layer.
+   * available pooling type: max, average
+   * @param input input values for this layer.
+   * @return output values for this layer.
+   */
+
+  @Override
+  public  Matrix feedForward(final Matrix input) {
+    switch (getLayerParameter().getPoolingType().toLowerCase()) {
```
beomyeol: Why don't we use `enum`? It will be more efficient than changing a string to lower case and comparing it with `"max"` and `"average"`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
seyounglee95: I left functions like `feedForwardMaxPooling()`, `feedForwardAveragePooling()` abstract functions since I didn't implement them. I found that if I made this class `final` I couldn't use `abstract` for functions in that class. Is there any other way to remove `abstract` keyword while maintaining abstract functions inside that class?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,144 @@
...
+/**
+ * Pooling layer.
+ *
+ * This layer is not learnable.
+ * This layer resizes input matrix spatially, using max pooling or average pooling.
+ * In a forward pass,
+ * max pooling picks maximum value in certain range ( kernalHeight * kernalWidth) and these values make up output.
+ * average pooling get average of values in certain range (kernalHeight * kernalWidth) and these values make up output.
+ * In a backward pass,
+ * each value of error matrix is the sum of elements in next error matrix
```
beomyeol: Are backpropagation algorithms same in both max pooling and average pooling?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  /**
+   * This function computes output size.
+   * input size: row * col
+   * output size: row' * col'
+   * row = (row − kernal_height) / stride + 1
+   * col = (col − kernal_width) / stride + 1
+   */
+
```
beomyeol: Please remove this redundant line.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,144 @@
...
+  public int[] setOutputShape() {
```
beomyeol: Should `setOutputShape()` be `private`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,144 @@
...
+    final int[] inputShape = getInputShape();
+    int[] computedShape;
```
beomyeol: `computedShape` should be `final`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
beomyeol: If you have an idea to implement `AveragePoolingLayer` and `MaxPoolingLayer` derived from this `PoolingLayer`, `PoolingLayer` can be an abstract class, leaving `feedForward()` and `backPropagate()` as abstract functions.
Otherwise, we can throw an exception in the body of `feedForwardMaxPooling()` and `feedForwardAveragePooling()`, for example, `RuntimeException("Not implemented")`. In the later case, we can remove `abstract` keyword from the class declaration and `feedFoward*Pooling()` functions. `feedFoward*Pooling()` functions should be `private` in this case.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,144 @@
...
+   * available pooling type: max, average
+   * @param input the input values for this layer.
+   * @param activation the output values.
+   * @param nextError the errors of the next layer - the one closer to the output layer.
+   * @return errors for this layer with the specified input value.
+   */
+
+  @Override
+  public Matrix backPropagate(final Matrix input, final Matrix activation, final Matrix nextError) {
+    switch (getLayerParameter().getPoolingType().toLowerCase()) {
```
beomyeol: Same as above, using `enum` will be better.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   * Computes errors for this pooling layer.
+   * available pooling type: max, average
+   * @param input the input values for this layer.
+   * @param activation the output values.
+   * @param nextError the errors of the next layer - the one closer to the output layer.
+   * @return errors for this layer with the specified input value.
+   */
+
```
beomyeol: Please remove this unnecessary empty line.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,144 @@
...
+    setLayerParameter(layerParameterInitializer.generateInitialParameter());
+  }
+
+  @Override
+  public int[] getOutputShape() {
+    return outputShape;
+  }
+
+  /**
+   * This function computes output size.
```
beomyeol: `shape` will be better than `size`. We should use consistent term.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
seyounglee95: I think backpropagation algorithms are same in both case because the point of backpropagation algorithm is to track where each value came from and add the error of that value to the errors of values it came from.

beomyeol:
@seyounglee95 This looks good. I finished my pass and left some comments. Please check them:)

seyounglee95:
@beomyeol  Thank you for your review. I fixed referring your comments. Please take a look.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/LayerConfigurationParameters.java
```diff
@@ -64,4 +64,27 @@
   @NamedParameter(doc = "loss function of loss layer", short_name = "lossFunc")
   public static final class LossFunction implements Name<String> {
   }
+
+  /**
+   * For pooling layers.
+   */
+  @NamedParameter(doc = "pooling type of pooling layer", short_name = "poolingTyp")
```
beomyeol: For this parameter, `short_name` seems unnecessary.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/PoolingLayerConfigurationBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+import edu.snu.dolphin.dnn.proto.NeuralNetworkProtos;
+import org.apache.reef.tang.Configuration;
+import org.apache.reef.tang.Tang;
+import org.apache.reef.util.Builder;
+
+/**
+ * Configuration builder for Pooling layer.
+ *
+ * The configuration that this builder generates is used to create a pooling layer instance.
+ * The generate configuration need to bind to bind the parameter for a layer input shape, to inject layer instance.
```
beomyeol: This description is weird.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/PoolingLayerConfigurationBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+    strideWidth = protoConf.getPoolingParam().getStrideWidth();
+    kernelHeight = protoConf.getPoolingParam().getKernelHeight();
+    kernelWidth = protoConf.getPoolingParam().getKernelWidth();
+    return this;
+  }
+
+  @Override
+  public synchronized Configuration build() {
+    return Tang.Factory.getTang().newConfigurationBuilder()
+        .bindNamedParameter(LayerConfigurationParameters.PoolingType.class, String.valueOf(poolingType))
```
beomyeol: `String.valueOf()` is unnecessary.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/PoolingLayerParameterInitializer.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.reef.tang.annotations.Parameter;
+
+import javax.inject.Inject;
+
+import static edu.snu.dolphin.dnn.util.NeuralNetworkUtils.shapeFromString;
+
+/**
+ * Pooling Layer parameter initializer.
+ *
+ * This initializer is for pooling layer which do not have layer parameter.
```
beomyeol: Please use plurals for pooling layer and layer parameter.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/PoolingLayerParameterInitializer.java
```diff
@@ -0,0 +1,98 @@
...
+  private final int index;
+  private final int[] inputShape;
+  private final int strideHeight;
+  private final int strideWidth;
+  private final int kernelHeight;
+  private final int kernelWidth;
+  private final LayerParameter emptyLayerParam;
+
+  @Inject
+  public PoolingLayerParameterInitializer(
```
beomyeol: Should be `private`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/PoolingLayerParameterInitializer.java
```diff
@@ -0,0 +1,98 @@
...
+   */
+  public int getIndex() {
+    return index;
+  }
+
+  /**
+   * This function computes output shape.
+   * input shape: row * col
+   * output shape: row' * col'
+   * row = (row − kernelHeight) / stride + 1
```
beomyeol: `row' =`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/PoolingLayerParameterInitializer.java
```diff
@@ -0,0 +1,98 @@
...
+   * col = (col − kernelWidth) / stride + 1
```
beomyeol: `col' = `?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/PoolingLayerParameterInitializer.java
```diff
@@ -0,0 +1,98 @@
...
+   */
```
beomyeol: Please complete a javadoc for this public function.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/PoolingLayerParameterInitializer.java
```diff
@@ -0,0 +1,98 @@
...
+      return computedShape;
+    case 2 :
+      computedShape = new int[2];
+      computedShape[0] = (inputShape[0] - kernelHeight) / strideHeight + 1;
+      computedShape[1] = (inputShape[1] - kernelWidth) / strideWidth + 1;
+      return computedShape;
+    default :
+      throw new IllegalArgumentException("Unsupported input dimension: " + Integer.toString(inputShape.length));
+    }
+  }
```
beomyeol: How about calculating the output shape in the constructor and saving it as a member variable?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,172 @@
...
+
+import javax.inject.Inject;
+
+/**
+ * Pooling layer.
+ *
+ * This layer is not learnable.
+ * This layer resizes input matrix spatially, using max pooling or average pooling.
+ * In a forward pass,
+ * max pooling picks maximum value in certain range (kernelHeight * kernelWidth) and these values make up output.
```
beomyeol: max pooling picks **the** maximum value

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,172 @@
...
+ * Average pooling gets average of values in certain range (kernelHeight * kernelWidth) and these values make up output.
```
beomyeol: gets **the** average of values

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,172 @@
...
+ * In a backward pass,
+ * each value of error matrix is the sum of elements in next error matrix
+ * on which had an impact in feedforward step.
```
beomyeol: no subject?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,172 @@
...
+ */
```
beomyeol: Please add the description that pooling layers work for only 1D and 2D inputs.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,172 @@
...
+public final class PoolingLayer extends LayerBase {
+
+  private enum PType {
```
beomyeol: `PoolingType` seems better than `PType`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,172 @@
...
+    average, max
```
beomyeol: Please use capital letters. `AVERAGE` and `MAX`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,172 @@
...
+  }
+  private final int[] outputShape;
+  private PType poolingType;
+  private int strideHeight;
+  private int strideWidth;
+  private int kernelHeight;
+  private int kernelWidth;
+  private Matrix trackMatrix;
+  private Function poolingFunctions;
```
beomyeol: Please add `final` keywords for member variables that will not be changed.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,172 @@
...
+  @Override
+  public int[] getOutputShape() {
+    return outputShape;
+  }
+
+  /**
+   * This function computes output shape.
+   * input shape: row * col
+   * output shape: row' * col'
+   * row = (row − kernelHeight) / stride + 1
```
beomyeol: `row' =`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,172 @@
...
+   * col = (col − kernelWidth) / stride + 1
```
beomyeol: `col' =`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,172 @@
...
+   */
```
beomyeol: Same as `getOutputShape()` in `PoolingLayerParameterInitializer`, please complete a javadoc for this.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,172 @@
...
+  private int[] setOutputShape() {
```
beomyeol: Please change the function name to `getOutputShape()`. It seems not like a setter.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  /**
+   * Computes output values for this pooling layer.
+   * available pooling type: max, average
+   * @param input input values for this layer.
+   * @return output values for this layer.
+   */
+  @Override
+  public  Matrix feedForward(final Matrix input) {
```
beomyeol: Two spaces between `public` and `Matrix`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,172 @@
...
+      return backPropagateAveragePooling(input, nextError);
+    default :
+      throw new IllegalArgumentException("Illegal pooling type: " + poolingType);
+    }
+  }
+
+  /** {@inheritDoc} */
+  @Override
+  public LayerParameter generateParameterGradient(final Matrix input, final Matrix error) {
+    throw new RuntimeException("This layer doesn't have parameter gradient");
```
beomyeol: Can you use the error message same as `ActivationLayer`'s one?

dolphin-dnn/src/main/proto/neural_network.proto
```diff
@@ -23,6 +23,14 @@ message FullyConnectedLayerConfiguration {
   required uint32 num_output = 4;
 }
 
+message PoolingLayerConfiguration {
+  required string pooling_type = 1;
```
beomyeol: Like Caffe, we can set `max` as a default value.

dolphin-dnn/src/main/proto/neural_network.proto
```diff
@@ -23,6 +23,14 @@ message FullyConnectedLayerConfiguration {
...
+  required uint32 stride_height = 2;
+  required uint32 stride_width = 3;
```
beomyeol: How about having default values for `stride_height` and `stride_width` as 1? If so, these can be optional.

beomyeol:
@seyounglee95 I finished another pass and left some comments.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
seyounglee95: There exists `PoolingType.class` so I couldn't use same name.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,172 @@
...
+
+  @Inject
+  private PoolingLayer(@Parameter(LayerIndex.class) final int index,
+                       @Parameter(LayerInputShape.class) final String inputShape,
+                       @Parameter(PoolingType.class) final String poolingType,
+                       @Parameter(StrideHeight.class) final int strideHeight,
+                       @Parameter(StrideWidth.class) final int strideWidth,
+                       @Parameter(KernelHeight.class) final int kernelHeight,
+                       @Parameter(KernelWidth.class) final int kernelWidth,
+                       final LayerParameterInitializer layerParameterInitializer) {
```
beomyeol: Please remove this unused variable.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
beomyeol: How about `PoolType`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
seyounglee95: If I change that function name to `getOutputShape()`, `getOutputShape()` would compute the same value again  whenever that function is called. How about just changing the function name to something like `computeOutputShape()` and keep `getOutputShape()` as now?

beomyeol: Sounds good.

seyounglee95:
@beomyeol Thank you for your review. I made some modifications. Please take a look.

dolphin-dnn/src/main/proto/neural_network.proto
```diff
@@ -23,6 +23,14 @@ message FullyConnectedLayerConfiguration {
   required uint32 num_output = 4;
 }
 
+message PoolingLayerConfiguration {
+  optional string pooling_type = 1;
```
beomyeol: We can set a default value for optional parameters as following.
`optional string pooling_type = 1 [default = "max"];`
FYI, here's the Caffe's protocol buffer definition file [[link]](https://github.com/BVLC/caffe/blob/master/src/caffe/proto/caffe.proto).

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/PoolingLayerConfigurationBuilder.java
```diff
@@ -0,0 +1,98 @@
...
+ */
+public final class PoolingLayerConfigurationBuilder implements Builder<Configuration> {
+
+  public static PoolingLayerConfigurationBuilder newConfigurationBuilder() {
+    return new PoolingLayerConfigurationBuilder();
+  }
+
+  private String poolingType = "MAX";
+  private int strideHeight = 1;
+  private int strideWidth = 1;
```
beomyeol: The default values for `poolingType`, `strideHeight`, and `strideWidth` should be set in `neural_network.proto`. Otherwise, we should check whether these parameters are set or not, which makes the code complicated.

seyounglee95: Then, I don't have to check `hasPoolingType()` or 'hasStride()`  in `fromProtoConfiguration()` if I set default values in `neural_network.proto`?

beomyeol: You don't have to check. The `get*()` returns the default values when they are not specified.

seyounglee95:
@beomyeol  I pushed a commit referring your comment. Please check it.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/PoolingLayerConfigurationBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+import edu.snu.dolphin.dnn.proto.NeuralNetworkProtos;
+import org.apache.reef.tang.Configuration;
+import org.apache.reef.tang.Tang;
+import org.apache.reef.util.Builder;
+
+/**
+ * Configuration builder for Pooling layer.
+ *
+ * The configuration that this builder generates is used to create a pooling layer instance.
+ * The generate configuration need to bind the parameter for a layer input shape, to inject layer instance.
```
beomyeol: The generate**d** configuration need**s** to bind ~, to inject **a** layer instance.
This grammatical error is also found in other layer configuration builders. Could you kindly fix all of them?

seyounglee95:
@beomyeol  I fixed comments. Thank you :D

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,169 @@
...
+ * In a backward pass,
+ * each value of error matrix is the sum of elements in next error matrix on which had an impact in feedforward step.
+ */
+public final class PoolingLayer extends LayerBase {
+
+  private enum PoolType {
+    AVERAGE, MAX
+  }
+  private final int[] outputShape;
+  private PoolType poolingType;
```
beomyeol: Please add `final`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,169 @@
...
+  private int strideHeight;
+  private int strideWidth;
+  private int kernelHeight;
+  private int kernelWidth;
```
beomyeol: Please add `final`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,169 @@
...
+public final class PoolingLayer extends LayerBase {
+
+  private enum PoolType {
+    AVERAGE, MAX
+  }
+  private final int[] outputShape;
+  private PoolType poolingType;
+  private int strideHeight;
+  private int strideWidth;
+  private int kernelHeight;
```
beomyeol: Please add `final`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,169 @@
...
+ */
+public final class PoolingLayer extends LayerBase {
+
+  private enum PoolType {
+    AVERAGE, MAX
+  }
+  private final int[] outputShape;
+  private PoolType poolingType;
+  private int strideHeight;
+  private int strideWidth;
```
beomyeol: Please add `final`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,169 @@
...
+ * each value of error matrix is the sum of elements in next error matrix on which had an impact in feedforward step.
+ */
+public final class PoolingLayer extends LayerBase {
+
+  private enum PoolType {
+    AVERAGE, MAX
+  }
+  private final int[] outputShape;
+  private PoolType poolingType;
+  private int strideHeight;
```
beomyeol: Please add `final`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * This layer is not learnable.
+ * This layer resizes input matrix spatially, using max pooling or average pooling.
+ * This layer works for only 1D and 2D inputs.
+ * In a forward pass,
+ * max pooling picks the maximum value in certain range (kernelHeight * kernelWidth) and these values make up output.
+ * Average pooling gets the average of values in certain range (kernelHeight * kernelWidth)
+ * and these values make up output.
+ * In a backward pass,
+ * each value of error matrix is the sum of elements in next error matrix on which had an impact in feedforward step.
```
beomyeol: Who did have an impact on next error matrices in forward step?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
seyounglee95: I'll fix this right away. I thought I fixed this, but I had some trouble handling git so the change was not saved. I'm sorry for bother you.

beomyeol: No problem :)

seyounglee95:
@beomyeol  Thank you for comments. I fixed declaration and comments.

beomyeol:
@seyounglee95 Looks good. I finished my pass. @jsjason, can you take a look?

jsjason:
@beomyeol @seyounglee95 Yes, I'll do a pass. I'll merge this right away (given that I don't see any problems), since this PR has been open for a long time.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/LayerConfigurationParameters.java
```diff
@@ -64,4 +64,27 @@
...
   }
+
+  /**
+   * For pooling layers.
+   */
+  @NamedParameter(doc = "pooling type of pooling layer")
+  public static final class PoolingType implements Name<String> {
+  }
+
+  @NamedParameter(doc = "stride height of pooling / convolutional layer", short_name = "strideH")
```
jsjason: @beomyeol Come to think of it, do we need short names for these parameters? Everything's being done by protobuf right now.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,169 @@
...
+
+  /**
+   * This function computes output shape.
+   * input shape: row * col
+   * output shape: row' * col'
+   * row' = (row − kernelHeight) / stride + 1
+   * col' = (col − kernelWidth) / stride + 1
+   * @return shape of output
+   */
+  private int[] computeOutputShape() {
```
jsjason: Why do we need this method here? We could receive an injection of `LayerParameterInitializer` (which would actually be `PoolingLayerParameterInitializer`) and call `LayerParameterInitializer.getOutputShape()`, instead of doing the same calculation again.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -0,0 +1,169 @@
...
+    this.kernelHeight = kernelHeight;
+    this.kernelWidth = kernelWidth;
+    this.outputShape = computeOutputShape();
+
+    try {
+      this.poolingType = PoolType.valueOf(poolingType);
+    } catch (final IllegalArgumentException illegalArgumentException) {
+      throw new IllegalArgumentException("Illegal pooling type: " + illegalArgumentException);
+    } catch (final NullPointerException nullPointerException) {
+      throw new NullPointerException("Null pointer exception while matching pooling type: " + nullPointerException);
```
jsjason: Instead of catching the exceptions and doing something relevant, this code just throws the same exceptions with a little change in the descriptions. Do we really need these try-catches?

jsjason:
@seyounglee95 Great. I'll merge this after you're done checking my comments.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/LayerConfigurationParameters.java
beomyeol: @jsjson You're right. We don't need short names. It will be more clear to remove them.

seyounglee95: Then we don't need short names for all parameters?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
seyounglee95: I thought these try-catches would be helpful for figuring out what's wrong with input data. And `ActivationWithLossLayer` also throws exception instead of doing something when error is caught, so I thought it would be better to use try-catches for `PoolingLayer` for code consistency. Is there any difference with try-catches in `ActivationWithLossLayer` and those in `PoolingLayer`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/conf/LayerConfigurationParameters.java
jsjason: Let's remove them altogether in another PR.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
jsjason: Actually, the exceptions in `ActivationWithLossLayer` need to be caught because they're not RuntimeExceptions, whereas these exceptions here do not need to be caught.

seyounglee95:
Thank you for review. Please take a look.

jsjason:
Great. I'll merge this.

PR #169, opened 2016-01-28
Pooling layer full implementation

seyounglee95:
This PR includes full implementation for pooling layer.  Now `PoolingLayerParameterInitializer.java` has function that checks the shape of output before being used in calculation. Also, 1D output would be transformed into 2D output for the convenience of calculation. In `PoolingLayer.java`,  concrete codes for `feedForwardMaxPooling()`, `feedForwardAveragePooling()`, `backPropagateMaxPooling()` and `backPropagateAveragePooling()` are added.

### DIFF
--- a/dolphin-dnn/README.md
+++ b/dolphin-dnn/README.md
@@ -79,7 +79,16 @@ parameter_provider {
 	* `random_seed`: the seed for generating random inital parameters.
 	* `num_output`: the number of outputs for this layer.
 
-**More types of layers such as convolutional layer and subsampling layer will be supported.**
+##### Pooling Layer
+* Layer Type: `Pooling`
+* Parameters (`PoolingLayerConfiguration pooling_param`)
+    * `pooling_type`[default="MAX"]: the type of pooling for this layer. Available types are MAX and AVERAGE.
+    * `stride_height`[default=1]: the height of stride that controls how new rows are allocated to compute.
+    * `stride_width`[default=1]: the width of stride that controls how new columns are allocated to compute.
+    * `kernel_height`: the height of kernel for this layer.
+    * `kernel_width`: the width of kernel for this layer.
+
+**More types of layers such as convolutional layer will be supported.**
 
 ##### Activation Layer
 * Layer type: `Activation`
--- a/dolphin-dnn/README.md
+++ b/dolphin-dnn/README.md
@@ -83,8 +83,8 @@ parameter_provider {
 * Layer type: `Pooling`
 * Parameters (`PoolingLayerConfiguration pooling_param`)
     * `pooling_type`[default="MAX"]: the type of pooling for this layer. Available types are MAX and AVERAGE.
-    * `stride_height`[default=1]: the height of stride that controls how new rows are allocated to compute.
-    * `stride_width`[default=1]: the width of stride that controls how new columns are allocated to compute.
+    * `stride_height`[default=1]: the interval at which pooling layers apply filter to inputs.
+    * `stride_width`[default=1]: the interval at which pooling layers apply filter to inputs.
     * `kernel_height`: the height of kernel for this layer.
     * `kernel_width`: the width of kernel for this layer.
 
--- a/dolphin-dnn/README.md
+++ b/dolphin-dnn/README.md
@@ -83,8 +83,8 @@ parameter_provider {
 * Layer type: `Pooling`
 * Parameters (`PoolingLayerConfiguration pooling_param`)
     * `pooling_type`[default="MAX"]: the type of pooling for this layer. Available types are MAX and AVERAGE.
-    * `stride_height`[default=1]: the interval at which pooling layers apply filter to inputs.
-    * `stride_width`[default=1]: the interval at which pooling layers apply filter to inputs.
+    * `stride_height`[default=1]: the interval at which pooling layers apply filters to inputs.
+    * `stride_width`[default=1]: the interval at which pooling layers apply filters to inputs.
     * `kernel_height`: the height of kernel for this layer.
     * `kernel_width`: the width of kernel for this layer.
 
--- a/dolphin-dnn/README.md
+++ b/dolphin-dnn/README.md
@@ -80,7 +80,7 @@ parameter_provider {
 	* `num_output`: the number of outputs for this layer.
 
 ##### Pooling Layer
-* Layer Type: `Pooling`
+* Layer type: `Pooling`
 * Parameters (`PoolingLayerConfiguration pooling_param`)
     * `pooling_type`[default="MAX"]: the type of pooling for this layer. Available types are MAX and AVERAGE.
     * `stride_height`[default=1]: the height of stride that controls how new rows are allocated to compute.
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/PoolingLayerParameterInitializer.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/PoolingLayerParameterInitializer.java
@@ -74,6 +74,17 @@ public final class PoolingLayerParameterInitializer implements LayerParameterIni
   }
 
   /**
+   * This function checks if stride is set proper for the input shape.
+   */
+  private void checkShape() {
+    if ((inputShape[0] - kernelHeight) % strideHeight != 0) {
+      throw new IllegalArgumentException("Stride height is not proper for input.");
+    } else if (inputShape.length == 2 && (inputShape[1] - kernelWidth) % strideWidth != 0) {
+      throw new IllegalArgumentException("Stride width is not proper for input.");
+    }
+  }
+
+  /**
    * This function computes output shape.
    * input shape: row * col
    * output shape: row' * col'
@@ -82,14 +93,14 @@ public final class PoolingLayerParameterInitializer implements LayerParameterIni
    * @return shape of output
    */
   private int[] computeOutputShape() {
-    final int[] computedShape;
+    final int[] computedShape = new int[2];
+    checkShape();
     switch (inputShape.length) {
     case 1:
-      computedShape = new int[1];
       computedShape[0] = (inputShape[0] - kernelHeight) / strideHeight + 1;
+      computedShape[1] = 1;
       return computedShape;
     case 2:
-      computedShape = new int[2];
       computedShape[0] = (inputShape[0] - kernelHeight) / strideHeight + 1;
       computedShape[1] = (inputShape[1] - kernelWidth) / strideWidth + 1;
       return computedShape;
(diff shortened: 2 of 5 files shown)


beomyeol:
@seyounglee95 We should add a unit test for pooling layers.

bgchun:
+1 for adding tests.

seyounglee95:
@beomyeol  Thank you for your review. I resolved some problem. Please take a look.

beomyeol:
@seyounglee95 I'll do a pass.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/PoolingLayerParameterInitializer.java
```diff
@@ -74,6 +74,17 @@ public int getIndex() {
   }
 
   /**
+   * This function checks if stride is set proper for the input shape.
```
beomyeol: set proper**ly**?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/PoolingLayerParameterInitializer.java
```diff
@@ -74,6 +74,17 @@ public int getIndex() {
...
+   */
+  private void checkShape() {
```
beomyeol: I am wondering if we cannot support this case. Caffe seems to work for this case. Can you check it?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layerparam/initializer/PoolingLayerParameterInitializer.java
```diff
@@ -82,20 +93,11 @@ public int getIndex() {
...
-      computedShape = new int[2];
-      computedShape[0] = (inputShape[0] - kernelHeight) / strideHeight + 1;
-      computedShape[1] = (inputShape[1] - kernelWidth) / strideWidth + 1;
-      return computedShape;
-    default:
-      throw new IllegalArgumentException("Unsupported input dimension: " + Integer.toString(inputShape.length));
-    }
+    final int[] computedShape = new int[2];
+    checkShape();
+    computedShape[0] = (inputShape[0] - kernelHeight) / strideHeight + 1;
```
beomyeol: I have a concern that we're missing outputs. For example, suppose that `inputShape[0]` is `6`, `kernelHeight` is `3`, and `strideHeight` is `2`. We compute only for the ranges [0, 2] and [2, 4], but we don't for [4, 5].  I know that the exception will be thrown in `checkShape()` for this case. However, shouldn't we support this? Caffe generates outputs for the range [4, 5]. Thus, the height of Caffe's output is `3`, not `2`.
Sorry that I missed this when reviewing last pull request.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -46,6 +47,8 @@
   private final int strideWidth;
   private final int kernelHeight;
   private final int kernelWidth;
+  private Matrix indexMatrix;
+  private MatrixFactory matrixFactory;
```
beomyeol: Please add `final`.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -76,19 +81,75 @@ public boolean isLearnable() {
...
 
+  /**
+   * Feedforward function for max pooling.
+   * @param input the input values for this layer.
+   * @return the output values for this layer.
+   */
   private Matrix feedForwardMaxPooling(final Matrix input) {
-    throw new RuntimeException("Not implemented");
+    final int[] inputShape = getInputShape();
+    final Matrix output = matrixFactory.create(outputShape[0] * outputShape[1], input.getColumns());
```
beomyeol: Why don't we use `NeuralNetworkUtils.getShapeLength(outputShape)` instead of `outputShape[0] * outputShape[1]`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -76,19 +81,75 @@ public boolean isLearnable() {
...
+    indexMatrix = matrixFactory.create(outputShape[0] * outputShape[1], input.getColumns());
```
beomyeol: The product of output shape dimensions can be calculated once and we can use it repeatedly.
In addition, we don't have to create a new `indexMatrix` every time. Why don't we generate `indexMatrix` in the constructor and just update its elements? If we do, `indexMatrix` should be `final`. This can remove unnecessary memory allocation.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
```diff
@@ -102,12 +163,54 @@ public Matrix feedForward(final Matrix input) {
...
+   */
   private Matrix backPropagateMaxPooling(final Matrix input, final Matrix nextError) {
-    throw new RuntimeException("Not implemented");
+    final Matrix error = matrixFactory.zeros(input.getRows(), input.getColumns());
+    for (int n = 0; n < input.getColumns(); ++n) {
+      for (int oh = 0; oh < outputShape[0]; ++oh) {
+        for (int ow = 0; ow < outputShape[1]; ++ow) {
+          //Add error to saved index.
+          final int index = oh * outputShape[1] + ow;
+          final float tempError = nextError.get(index, n) + error.get(index, n);
```
beomyeol: I'm wondering if `error.get(index, n)` is correct. Is it `error.get((int) indexMatrix.get(index, n), n)`?

dolphin-dnn/src/test/java/edu/snu/dolphin/dnn/layers/PoolingLayerTest.java
```diff
@@ -0,0 +1,159 @@
...
+import org.apache.reef.tang.Tang;
+import org.apache.reef.tang.exceptions.InjectionException;
+import org.junit.Before;
+import org.junit.Test;
+import edu.snu.dolphin.dnn.conf.LayerConfigurationParameters.*;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Test class for pooling layer.
+ */
```
beomyeol: We should have add `@FixMethodOrder(MethodSorters.NAME_ASCENDING)` for test success.
`testMaxPoolingActivation()` should be executed before `testMaxPoolingBackPropagate()` is executed. Here is the link about JUnit's test execution order [link](https://github.com/junit-team/junit/wiki/Test-execution-order).

dolphin-dnn/src/test/java/edu/snu/dolphin/dnn/layers/PoolingLayerTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+  @Before
+  public void setup() throws InjectionException {
+    final Configuration layerConfMax = Tang.Factory.getTang().newConfigurationBuilder()
+        .bindNamedParameter(LayerIndex.class, String.valueOf(0))
+        .bindNamedParameter(LayerInputShape.class, "3,3")
+        .bindImplementation(MatrixFactory.class, MatrixJBLASFactory.class)
+        .build();
+
+    final Configuration layerConfAverage = Tang.Factory.getTang().newConfigurationBuilder()
```
beomyeol: If `layerConfMax` and `layerConfAverage` are same, why don't we have a single layer configuration  `layerConf` and use it for injecting both max pooling layer and average pooling layer?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
seyounglee95: Oh, my mistake. `error.get((int) indexMatrix.get(index, n), n)` is right.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
seyounglee95: I think that's not possible since we don't know the number of input images when the class is newly created but we need to `input.getColumns()` to create `indexMatrix`.

beomyeol:
@seyounglee95 I finished my pass and left some comments.
I also have a concern that `IllegalArgumentException` will be thrown when users do not specify `pooling_type` as uppercase. We should convert `poolingType` to uppercase in the constructor of `PoolingLayer` before `PoolType.valueOf()`.

beomyeol:
@seyounglee95 Could you add the description for pooling layers in dolphin-dnn README?

seyounglee95:
@beomyeol Thank you for your comments. I added the description in README and resolved the problem about the output shape that is not divided by stride. Please take a look.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/PoolingLayer.java
beomyeol: I see. Just leave it as it is.

beomyeol:
Even though we describe that pooling layers only support 2D input, it will be better to throw an explicit exception indicating that users try to use not supported operation. What do you think of this?

dolphin-dnn/README.md
```diff
@@ -79,7 +79,16 @@ parameter_provider {
 	* `random_seed`: the seed for generating random inital parameters.
 	* `num_output`: the number of outputs for this layer.
 
-**More types of layers such as convolutional layer and subsampling layer will be supported.**
+##### Pooling Layer
+* Layer type: `Pooling`
+* Parameters (`PoolingLayerConfiguration pooling_param`)
+    * `pooling_type`[default="MAX"]: the type of pooling for this layer. Available types are MAX and AVERAGE.
+    * `stride_height`[default=1]: the height of stride that controls how new rows are allocated to compute.
```
beomyeol: It is not clear that a stride controls how new rows are allocated to compute.
Does a stride represent the interval at which pooling layers apply filter to inputs?

dolphin-dnn/README.md
```diff
@@ -79,7 +79,16 @@ parameter_provider {
...
+    * `stride_width`[default=1]: the width of stride that controls how new columns are allocated to compute.
```
beomyeol: It is not clear that a stride controls how new rows are allocated to compute.
Does a stride represent the interval at which pooling layers apply filter to inputs?

beomyeol:
@seyounglee95 This looks awesome! I'm merging this. :)

PR #174, opened 2016-02-18
Convolutional layer full implementation

seyounglee95:
This PR includes full implementation of convolutional layers. `im2col()` and `col2im()` are implemented to facilitate matrix multiplication in `feedForward()`, `backPropagate()` and `generateParameterGradient()`. Convolutional layer unit test class is added, too. This layer supports only 2D inputs and support for 3D inputs will be implemented in later PR.

### DIFF
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/ConvolutionalLayer.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/ConvolutionalLayer.java
@@ -132,10 +132,8 @@ public final class ConvolutionalLayer extends LayerBase {
             int iw = kw - paddingWidth;
             for (int ow = 0; ow < outputShape[1]; ++ow) {
               if (iw >= 0 && iw < inputShape[1]) {
-                final int colh = colIndex / kernelSize;
-                final int colw = colIndex % kernelSize;
                 final int inputIndex = ih * inputShape[1] + iw;
-                final float newValue = col.get(colh, colw) + im.get(inputIndex);
+                final float newValue = col.get(colIndex) + im.get(inputIndex);
                 im.put(inputIndex, newValue);
               }
               colIndex++;
@@ -176,7 +174,9 @@ public final class ConvolutionalLayer extends LayerBase {
   public Matrix backPropagate(final Matrix input, final Matrix activation, final Matrix nextError) {
     final Matrix error = matrixFactory.create(input.getRows(), input.getColumns());
     for (int n = 0; n < input.getColumns(); ++n) {
-      final Matrix im = col2im(nextError.getColumn(n).mmul(getLayerParameter().getWeightParam()));
+      final Matrix colVec = nextError.getColumn(n);
+      final Matrix col = getLayerParameter().getWeightParam().transpose().mmul(colVec.reshape(1, colVec.getLength()));
+      final Matrix im = col2im(col);
       error.putColumn(n, im);
     }
     return error;
--- a/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/ConvolutionalLayer.java
+++ b/dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/ConvolutionalLayer.java
@@ -90,7 +90,7 @@ public final class ConvolutionalLayer extends LayerBase {
   private Matrix im2col(final int imageIndex, final Matrix input) {
     final int[] inputShape = getInputShape();
     final Matrix col =
-        matrixFactory.zeros(NeuralNetworkUtils.getShapeLength(outputShape), kernelHeight * kernelWidth);
+        matrixFactory.zeros(kernelHeight * kernelWidth, NeuralNetworkUtils.getShapeLength(outputShape));
     for (int kh = 0; kh < kernelHeight; ++kh) {
       for (int kw = 0; kw < kernelWidth; ++kw) {
         int ih = kh - paddingHeight;
@@ -99,8 +99,8 @@ public final class ConvolutionalLayer extends LayerBase {
             int iw = kw - paddingWidth;
             for (int ow = 0; ow < outputShape[1]; ++ow) {
               if (iw >= 0 && iw < inputShape[1]) {
-                col.put(oh * outputShape[1] + ow,
-                    kh * kernelHeight + kw, input.get(ih * inputShape[1] + iw, imageIndex));
+                col.put(kh * kernelWidth + kw, oh * outputShape[1] + ow,
+                    input.get(ih * inputShape[1] + iw, imageIndex));
               }
               iw += strideWidth;
             }
@@ -135,8 +135,8 @@ public final class ConvolutionalLayer extends LayerBase {
                 final int colh = colIndex / kernelSize;
                 final int colw = colIndex - colh * kernelSize;
                 final int inputIndex = ih * inputShape[1] + iw;
-                final float newValue = col.get(colh, colw) + im.get(inputIndex, 0);
-                im.put(inputIndex, 0, newValue);
+                final float newValue = col.get(colh, colw) + im.get(inputIndex);
+                im.put(inputIndex, newValue);
               }
               colIndex++;
               iw += strideWidth;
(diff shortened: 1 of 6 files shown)


beomyeol:
@seyounglee95 I'll do a pass.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/ConvolutionalLayer.java
```diff
@@ -77,13 +82,87 @@ public boolean isLearnable() {
   }
 
   /**
+   * Transform the given image to column form to facilitate matrix multiplication.
```
beomyeol: This rearranges from the image into rows, doesn't it? If so, the method name `im2col` is misleading.
We can change `im2col` to `im2row`. However, it may not be familiar to people who use MATLAB or GNU Octave which provide an embedded `im2col` function.
Thus, we can change this function to transform from the image to columns and `col.mmul(getLayerParameter().getWeightParam()` to `getLayerParameter().getWeightParam().mmul(col)`. How do you think about this?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/ConvolutionalLayer.java
```diff
@@ -77,13 +82,87 @@ public boolean isLearnable() {
...
+    for (int kh = 0; kh < kernelHeight; ++kh) {
+      for (int kw = 0; kw < kernelWidth; ++kw) {
+        int ih = kh - paddingHeight;
+        for (int oh = 0; oh < outputShape[0]; ++oh) {
+          if (ih >= 0 && ih < inputShape[0]) {
+            int iw = kw - paddingWidth;
+            for (int ow = 0; ow < outputShape[1]; ++ow) {
+              if (iw >= 0 && iw < inputShape[1]) {
+                col.put(oh * outputShape[1] + ow,
+                    kh * kernelHeight + kw, input.get(ih * inputShape[1] + iw, imageIndex));
```
beomyeol: kh * ~~kernelHeight~~**kernelWidth** + kw?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/ConvolutionalLayer.java
```diff
@@ -77,13 +82,87 @@ public boolean isLearnable() {
...
+          if (ih < 0 || ih >= inputShape[0]) {
+            colIndex += outputShape[1];
+          } else {
+            int iw = kw - paddingWidth;
+            for (int ow = 0; ow < outputShape[1]; ++ow) {
+              if (iw >= 0 && iw < inputShape[1]) {
+                final int colh = colIndex / kernelSize;
+                final int colw = colIndex - colh * kernelSize;
+                final int inputIndex = ih * inputShape[1] + iw;
+                final float newValue = col.get(colh, colw) + im.get(inputIndex, 0);
```
beomyeol: Why don't we use `im.get(inputIndex)` instead of `im.get(inputIndex, 0)`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/ConvolutionalLayer.java
```diff
@@ -77,13 +82,87 @@ public boolean isLearnable() {
...
+                im.put(inputIndex, 0, newValue);
```
beomyeol: Why don't we use `im.put(inputIndex, newValue)` instead of `im.put(inputIndex, 0, newValue)`?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/ConvolutionalLayer.java
```diff
@@ -95,12 +174,25 @@ public Matrix feedForward(final Matrix input) {
...
+      error.putColumn(n, im);
+    }
+    return error;
   }
 
   /** {@inheritDoc} */
   @Override
   public LayerParameter generateParameterGradient(final Matrix input, final Matrix error) {
-    throw new RuntimeException("Not Implemented");
+    final Matrix weightDiff = matrixFactory.create(kernelHeight * kernelWidth, 1);
```
beomyeol: Since we have not used a term "diff" for gradients, it would be better to use another term like "gradient". How about `weightGradient` instead of `weightDiff`?

dolphin-dnn/src/test/java/edu/snu/dolphin/dnn/layers/ConvolutionalLayerTest.java
```diff
@@ -0,0 +1,228 @@
...
+import org.junit.Test;
+import edu.snu.dolphin.dnn.conf.LayerConfigurationParameters.*;
+import org.junit.runners.MethodSorters;
+
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Test class for convolutional layer.
+ */
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
```
beomyeol: We don't have to specify the test order because `ConvolutionLayer` is not stateful like `PoolingLayer`.

dolphin-dnn/src/test/java/edu/snu/dolphin/dnn/layers/ConvolutionalLayerTest.java
```diff
@@ -0,0 +1,228 @@
...
+public class ConvolutionalLayerTest {
+  private static MatrixFactory matrixFactory;
```
beomyeol: This is a minor one. Would it be better to add a newline like others?

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/ConvolutionalLayer.java
```diff
@@ -77,13 +82,87 @@ public boolean isLearnable() {
...
+   * Transform the given image to column form to facilitate matrix multiplication.
+   * @param imageIndex the index of the image in the input matrix.
+   * @param input input values for this layer.
+   * @return the converted column.
+   */
+  private Matrix im2col(final int imageIndex, final Matrix input) {
+    final int[] inputShape = getInputShape();
+    final Matrix col =
+        matrixFactory.zeros(NeuralNetworkUtils.getShapeLength(outputShape), kernelHeight * kernelWidth);
+    for (int kh = 0; kh < kernelHeight; ++kh) {
```
beomyeol: This part seems to have the opportunity to to be optimized.
For example, since our matrix implementation uses column-major order, it might be more efficient to use the loop for `kh` as the innermost loop, which can increase cache hit rate for `input`.

seyounglee95:
@beomyeol  I made new branch `issue-171-convolutional_layer_full_implementation_temp`. Please check this.

dolphin-dnn/src/main/java/edu/snu/dolphin/dnn/layers/ConvolutionalLayer.java
beomyeol: @seyounglee95 and I discussed this issue in an offline meeting, Caffe shows better performance with this implementation. However while Caffe uses row-major order, we use column-major order. This difference may influence performance regarding to memory access pattern.
We will take this issue into account later after considering difference implementations with a micro-benchmark.

seyounglee95:
@beomyeol  Thank you for your review. I pushed a commit. Please take a look.

dolphin-dnn/src/test/java/edu/snu/dolphin/dnn/layers/ConvolutionalLayerTest.java
```diff
@@ -0,0 +1,226 @@
+/*
+ * Copyright (C) 2015 Seoul National University
```
beomyeol: This should be `2016`.

dolphin-dnn/src/test/java/edu/snu/dolphin/dnn/layers/ConvolutionalLayerTest.java
```diff
@@ -0,0 +1,226 @@
...
+        .getBiasParam().compare(convolutionalLayerParams.getBiasParam(), TOLERANCE));
+    assertTrue(expectedConvolutionalLayerParams
+        .getWeightParam().compare(convolutionalLayerParams.getWeightParam(), TOLERANCE));
+  }
+
+  @Test
+  public void testConvolutionalWithPaddingGradient() {
+    final LayerParameter convolutionalLayerParams =
+        convolutionalWithPaddingLayer.generateParameterGradient(input, nextErrorWithPadding);
+    assertTrue(expectedConvolutionalLayerWithPaddingParams
```
beomyeol: Checking the equality of two layer parameters is used several times. Why don't we create a method for this?

beomyeol:
@seyounglee95 This looks good. However, there are a couple of minor issues. I'll fix them and merging this.

PR #177, opened 2016-03-02
[DOLPHIN-176] Create partitioned, queued, single-node parameter server

### LINKED ISSUE #176: Create partitioned, queued, single-node parameter server
We should improve a single node parameter server. The high-level design will consist of partitions based on the hash of the key, with a queue, kvStore, and thread per partition.

The design is meant to improve on some of the issues with the current single-node parameter server.

* It will not miss updates (#152)
* It will decouple PS work from the NCS threads (#157)

This issue does not cover the client (Worker) side.

dafrista:
Closes #176

- Adds a new Parameter Server, PartitionedParameterServer. The server supports atomic, in-order processing of push and pull operations, using multiple threads.
- Renames the existing "Single Node" Server as ConcurrentParameterServer.

I've done a simple performance comparison between PartitionedPS and ConcurrentPS by running the unit tests (1million reads + 1million writes) on my local laptop. The PartitionedPS has worse performance. Note that the ConcurrentPS can lose updates.
- ConcurrentPS: Avg 1.5s
- PartitionedPS: Avg 2s

Also, the unit test does not include hash computation. A version of the test (not committed) that uses MurmurHash had an Avg runtime of 3s. However, this test also requires the key to be copied into a byte array, which we avoid in the actual implementation by using the original serialized value from the message.

### DIFF
--- a/bin/run_ps.sh
+++ b/bin/run_ps.sh
@@ -1,0 +1,30 @@
+#!/bin/sh
+# Copyright (C) 2016 Seoul National University
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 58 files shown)


jsjason:
I did a pass and everything seems great. The new tests work as intended, and the existing dnn w/ ps example works well too. I'm merging this.

PR #178, opened 2016-03-03
[DOLPHIN-154] Create a Partitioned PS Worker

### LINKED ISSUE #154: Use external libraries like Guava or Guice to simplify code
There are a few points in our parameter server code that are overly complicated, such as `SingleNodeParameterWorker#pull()`. We can just use external libraries for better performance as well as code tidiness.

dafrista:
Closes #154.

This creates a Partitioned PS Worker.
   * A remotely read pull remains in the local cache for a duration of expireTimeout.
   * Pushes are applied locally while the parameter is cached.
   * The per partition single queue-and-thread, combined with the server, provides a guarantee that all previous local pushes are applied to a pull, if it is locally cached.

On local tests, the performance is comparable to ConcurrentParameterWorker.

There is a bug and workaround in this code:

We are observing some lost messages when contexts are immediately closed after the task completes in the example. It appears messages buffered in NCS are not being flushed before context close, but this has to be investigated further. The current workaround in the example is to not close contexts immediately, but this must be resolved before we move to elastically removing contexts.

(This should not be a Worker problem. The `ContextStopHandler` waits for messages to be flushed from the Worker queue.)

### DIFF
--- a/dolphin-ps/src/main/java/edu/snu/dolphin/ps/driver/ParameterServerDriver.java
+++ b/dolphin-ps/src/main/java/edu/snu/dolphin/ps/driver/ParameterServerDriver.java
@@ -114,6 +114,7 @@ public final class ParameterServerDriver {
         .newConfigurationBuilder(
             codecConfiguration,
             psManager.getWorkerServiceConfiguration(),
+            updaterConfiguration,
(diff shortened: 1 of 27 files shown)


jsjason:
Checked the code and tests, all looks great.

PR #180, opened 2016-03-23
Use Apache REEF 0.14.

dongjoon-hyun:
Apache REEF 0.14 is released at the maven central.
We had better use 0.14 instead of 0.14-SNAPSHOT.


jsjason:
Thanks! Hopefully, we won't need to use 0.15.0-SNAPSHOT for a while.

dongjoon-hyun:
Thank you for merging, @jsjason ! :)

PR #181, opened 2016-03-23
[DOLPHIN-131] Add static multi-node parameter server

### LINKED ISSUE #131: Multi-node parameter server
A parameter server that consists of only one node is very likely to be a computation bottleneck, and can be viewed as a single point of failure that makes the whole job fail when a server fault happens. We need to implement a parameter server that is backed by multiple nodes.

beomyeol:
Closes #131 

Adds a multi-node parameter server that is statically configured based on the number of servers. I have tested this on my laptop with the ps example.


jsjason:
@beomyeol I checked the code and it looks great, as discussed. PartitionedPSExampleREEF works well too. I'm merging this.

